PR #3, opened 2021-11-23
Telco 87 telco 88 additional orc8r nginx services

Gmerold:


orchestrator-bundle/bundle-local.yaml
```diff
@@ -4,37 +4,37 @@ applications:
...
     trust: true
     resources:
       magma-nms-magmalte-image: docker.artifactory.magmacore.org/magmalte:1.6.0
-  magma-nms-nginx-proxy:
+  nms-nginx-proxy:
     charm: ./nms-nginx-proxy-operator/nms-nginx-proxy.charm
     scale: 1
     trust: true
     resources:
       magma-nms-nginx-proxy-image: nginx:latest
```
sanchezfdezjavier, line 18: It might be interesting to freeze the `nginx` image version. We did the same for the new `prometheus-edge-hub-operator` charm and seems like a smart move. Future versions of Nginx might break things. What do you think?

Gmerold: Normally I'd say yes, but in this case I'm afraid it may lead to getting out of sync with Magma (or additional effort needed to test and update nginx version every time new version is released).
As long as Magma uses latest, I think so should we.

sanchezfdezjavier: Good point, resolved 👍🏼

sanchezfdezjavier (commented):
Good job, it was quite big!

sanchezfdezjavier (approved):

PR #4, opened 2021-11-24
Telco 92 integration tests

Gmerold:


.github/workflows/orchestrator.nms.magmalte.yml
```diff
@@ -0,0 +1,50 @@
+name: CI
+
+on:
+  push:
+    paths:
+    - 'orchestrator-bundle/nms-magmalte-operator/**'
```
sanchezfdezjavier: Shouldn't the `-` be intended?
```suggestion
    paths:
      - 'orchestrator-bundle/nms-magmalte-operator/**'
```

sanchezfdezjavier: Actually, I think this applies to a lot of files. Idk know if this thing would throw a YAML error, but might be worth checking

Gmerold: You're right, over half of the yamls had that wrong... yet worked somehow... magic :D
Anyway, according to the docs, it should be indented.
Fixed.

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class TestNmsMagmaLTE:
+    @pytest.fixture(scope="module")
+    async def setup(self, ops_test):
+        await self._deploy_postgresql(ops_test)
+        await self._deploy_orc8r_certifier(ops_test)
+
+    @pytest.mark.abort_on_fail
+    async def test_build_and_deploy(self, ops_test, setup):
```
sanchezfdezjavier, line 29: There's no reference to the `setup` parameter within the function. Should it be dropped?

Gmerold: It's not a parameter; it's a reference to a fixture (defined above).

sanchezfdezjavier: This is repeated in more `test_integration.py` files

sanchezfdezjavier (approved):

PR #7, opened 2021-12-07
Implements orc8r bundle

gruyaume:


orchestrator-bundle/orc8r-bundle/README.md
```diff
@@ -0,0 +1,101 @@
...
+### Microk8s
+On your Ubuntu VM, install microk8s using snap:
+
+```bash
+sudo snap install microk8s --classic
+```
+Update your user's permission to be added to the microk8s group
+
+```bash
+sudo usermod -a -G microk8s ubuntu
```
sanchezfdezjavier: Do we want to hardcode the `username`? What do you think about the following?
```suggestion
sudo usermod -a -G microk8s $USER
```

orchestrator-bundle/orc8r-bundle/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Juju
+Install Juju using snap:
+
+```bash
+sudo snap install juju --classic
+```
+
+Create a Juju controller which is the management node of a Juju cloud environment. This will create
+a new
```
sanchezfdezjavier: I think this line is cropped

orchestrator-bundle/orc8r-bundle/bundle.yaml
```diff
@@ -0,0 +1,268 @@
...
+  configure and monitor the wireless network securely. The metrics acquired through the platform 
+  allows you to see the analytics and traffic flows of the wireless users through the Magma web UI.
+applications:
+  nms-magmalte:
+    charm: magma-nms-magmalte
+    series: kubernetes
+    channel: edge
+    revision: 3
+    resources:
+      magma-nms-magmalte-image: 1
```
Gmerold, line 14: Is this correct?
I couldn't find any reference, but shouldn't this be pointing to the docker image?

orchestrator-bundle/CONTRIBUTING.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Developing and testing
+Testing for each charm is done the same way. First `cd` into the charm directory and then use 
+`tox` like so:
+```shell
+tox -e lint      # code style
+tox -e static    # static analysis
+tox -e unit      # unit tests
+```
+
```
sanchezfdezjavier, line 11: I think this sometimes helps for newbies like me. Let me know what you think
```suggestion
**NOTE:** If you don't have `tox` installed yet, just run: `pip3 install tox`
```

orchestrator-bundle/orc8r-bundle/README.md
```diff
@@ -0,0 +1,101 @@
...
+## Pre-requisites
+This bundle of charms requires the following:
+- Ubuntu (20.04)
+- Microk8s (v1.22.4)
+- Juju (2.9.21)
+
+### Ubuntu
+For installing Ubuntu, please follow the [official documentation](https://releases.ubuntu.com/20.04/).
+
+### Microk8s
```
Gmerold: Perhpas we should link official installation instructions for Microk8s and Juju as well (same as for Ubuntu)? If anything changes in that matter, we won't have to update our docs.

orchestrator-bundle/CONTRIBUTING.md
```diff
@@ -0,0 +1,128 @@
...
+## Integration tests
+To run the integration tests suite, run the following commands:
+```bash
+tox -e integration
+```
+
+## Build
+Building and publishing charms is done using charmcraft (official documentation
+[here](https://juju.is/docs/sdk/publishing)). You can install charmcraft using `snap`:
```
sanchezfdezjavier, line 29: Having two `charmcraft` installation statements that "apparently" do the same, might be confusing. We may consider explaining why we're doing this like so. What do you think?
```suggestion
[here](https://juju.is/docs/sdk/publishing)). You need to install the latest version of charmcraft to use the publishing feature, as it is not available in the stable one. You can install it using `snap`:
```

orchestrator-bundle/CONTRIBUTING.md
```diff
@@ -0,0 +1,128 @@
...
+### Bundle
+
+Building charms is done using [charmcraft](https://github.com/canonical/charmcraft). To install 
+charmcraft:
+```bash
+sudo snap install charmcraft --classic
+```
+Initialize LXD:
+
+```bash
```
sanchezfdezjavier: I think that LXD already comes preinstalled with Ubuntu, but I leave this here just in case:
```suggestion
```bash
sudo snap install lxd --classic
```

orchestrator-bundle/orc8r-bundle/bundle.yaml
sanchezfdezjavier: That might be because the images are already uploaded to charmhub, that was my assumption

gruyaume: This bundle.yaml was mostly automatically generated using `juju export-bundle` and yes it works. Here we are simply pointing to the revision of the image we already have uploaded when uploading the charm and its image.

orchestrator-bundle/CONTRIBUTING.md
gruyaume: It is already installed with ubuntu 20.04 so I wouldn't add this comment. Using Ubuntu 20.04 is already a pre-requisite.

orchestrator-bundle/CONTRIBUTING.md
gruyaume: Yeah I didn't realize we had it twice, I removed the second one.

orchestrator-bundle/CONTRIBUTING.md
gruyaume: Fixed

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: Fixed

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: Fixed

orchestrator-bundle/orc8r-bundle/bundle.yaml
Gmerold: Got it, thanks!

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: Fixed

Gmerold (approved):

PR #9, opened 2021-12-17
Implements library for common code

gruyaume:
Implements library for common code
- New charm is created: https://charmhub.io/magma-orc8r-libs
- New library is created in this charm: orc8r_base
- Most of orc8r services are similar and this library contains the code that is in common. This will ease maintenance as changes will only have to be done at one place.
- 2 charms (orc8r-ha and orc8r-directoryd) make use of this new library


orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -0,0 +1,221 @@
...
+The constructor simply takes the following:
+- Reference to the parent charm (CharmBase)
+- The service name (str)
+- The startup command (str)
+- Whether a database relation is needed (bool)
+- The pebble Ready event method (ops.CharmEvent)
+
+## Getting Started
+To get started using the library, you just need to fetch the library using `charmcraft`. **Note
+that you also need to add `ops-lib-pgsql` to your charm's `requirements.txt`.**
```
jnsgruk: Hmm. This `ops-lib-pgsql` dependency seems a little broad. Surely not *all* of your charms need the PostgreSQL lib? I'd maybe pull some of the PostgreSQL stuff into a separate lib?

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -0,0 +1,221 @@
...
+            "/var/opt/magma/bin/directoryd "
+            "-logtostderr=true "
+            "-v=0"
+        )
+        self._orc8r_base = Orc8rBase(
+            self,
+            service_name="magma-orc8r-directoryd",
+            startup_command=startup_command,
+            database_relation_needed=True,
+            pebble_ready_event=self.on.magma_orc8r_directoryd_pebble_ready
```
jnsgruk: I think you can drop this from your the constructor in most cases and leave it as optional -- you should be able to derive which the pebble-ready event is from the `on` attribute of the charm you're passing in? The only thing that's not super clear to me is what happens when there are two containers in the pod?

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
jnsgruk: Following on from this having read the actual pebble-ready handler later on in this code -- note that you can bind multiple event handlers to one event, so this library could do what it needed, and then library that deals with postgres could *also* bind a handler to the same pebble-ready if required.

jnsgruk (commented):
Love the idea, couple of niggles to discuss! :)

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -0,0 +1,221 @@
...
+    def __init__(
+        self,
+        charm: CharmBase,
+        service_name: str,
+        startup_command: str,
+        pebble_ready_event: CharmEvents,
+        database_relation_needed: bool = False,
+        additional_environment_variables: dict = None,
+    ):
+        super().__init__(charm, "ocr8r-base")
```
sanchezfdezjavier:
```suggestion
        super().__init__(charm, "orc8r-base")
```

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -0,0 +1,221 @@
...
+          in the relation event.
+        """
+        db_connection_string = event.master
+        if self.charm.unit.is_leader() and db_connection_string is not None:
+            event.database = self.DB_NAME
+        elif event.database != self.DB_NAME or db_connection_string is None:
+            event.defer()
+            return
+
+    def _check_db_relation_has_been_established(self):
```
sanchezfdezjavier: How about renaming the method name to `_db_relation_established`. IMHO is more concise and readable in the `if` statements. What do you think?

sanchezfdezjavier (commented):
I've really liked the idea, all the `orc8r` charms felt the same. Now it's gonna be much easier to differentiate them. Thanks for adding this!

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: Yes we could have a different lib for pgsql but they'd be very similar. We wouldn't want to have both libraries be actuated at the same event for pebble ready because the `basic` one would go ahead and configure the pebble layer while the pgsql one would wait until the relation with the db is ready to start it (which is needed).

So the other option is to have 2 libraries (orcr8r_base and orc8r_base_with_db). The issue is that they would be very similar but the nice thing is that they'd be no `if` statement in them.

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: What would be the generic attribute for pebble_ready? While there is a `on.install`, I see no such thing as `on.pebble_ready` so that's why each charm had to specify its own attribute.

jnsgruk: It wouldn't be so generic, you'd need to inspect the `on` attribute and do some clever filtering on it's attributes.

The other option is: you're passing in the charm, which means you can get the name of the containers that are specified in the charm's metadata, and the event will always be `<container_name>_pebble_ready` :)

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
jnsgruk: Hmm. I still think including the library everywhere is not ideal. How about this:

- Have a library with the base and a library that *just* deals with the database relation
- In the base library, you can interpret the charm's metadata, and behave slightly differently if the charm specifies a relation with the `pgsql` interface?

This would mean that the base library would need to check some attribute of the charm, perhaps stored in the unit section of the peer relation, before creating pebble layers etc. The db library could be the thing that sets that initial value once the relation is set up?

gruyaume: You give me complexity, I give you duplication. :laughing:

gruyaume: One of those libraries would have to look inside the relationship with the db to bring out information that'll be used in the environment variables in the pebble layer. If it's in the pgsql lib (which makes more sense imo), we need a way to pass this info to the base library. If it's in the base library, this library will still need to have `if` conditions to know when to look there and it would contain logic around the postgres stuff which isn't ideal.

Here's what I see as a problem:
- We will have 2 libs that share the role of configuring the pebble layer and having the logic spread out between the two. The pg lib would be dependant on the base lib to do its job and the base lib would contain logic around the way that it expects pg to behave. It makes the code more complicated and less easy to refactor with the only benefit of removing ~20 lines of code duplication.

What I would propose is the following:
- Let's have 2 libs (orc8r_base and orc8r_with_db) that are very similar but 100% independent. The user, when creating a new charm, would simply choose between the 2 based on whether or not it is necessary to have a db connection for this charm (instead of having to choose between fetching 2 libs or just 1).

jnsgruk: FTR - we discussed (out of band) this and we're going with the proposal made by @gruyaume for now -- especially while the PostgreSQL charms are in flux, and the libraries may change anyway.

gruyaume: Fixed with a new lib for orc8r_base_db

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: Fixed

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: Fixed

orchestrator-bundle/orc8r-directoryd-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: Fixed

sanchezfdezjavier (approved):

PR #12, opened 2022-01-28
Charms now leverage orc8r_base and orc8r_base_db libraries

gruyaume:
- Charms now leverage orc8r_base and orc8r_base_db libraries
- Fixes CI names


sanchezfdezjavier (approved):
Thanks for that Guillaume! LGTM

PR #13, opened 2022-02-01
Adds edge releasing to CI

sanchezfdezjavier:
This PR introduces the possibility of automatically uploading the charms and bundle from the CI

- The workflows are triggered automatically on every `push` to the `main` branch if the target charm or bundle has been modified
- If we want to push to specific `tracks`, we'll need to create a branch with a name as in the second row of the table. And the `track` needs to be first created manually.
- If a new charm/bundle is created we'll need to create its corresponding workflow and upload it manually first
- For this PR the `pull_request` `Event Type` is disabled

Screenshot 2022-02-01 at 17 27 33


.github/workflows/orchestrator.libs.charm.upload.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    name: Charmhub upload orc8r-libs
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: canonical/charmhub-upload-action@0.2.1
+        with:
+          credentials: "${{ secrets.CHARMCRAFT_AUTH }}"
+          charm-path: ./orchestrator-bundle/orc8r-libs
+          upload-image: "false"
```
gruyaume, line 21: Is there an action that supports publishing libs?

We have the 2 following libs that we publish:
```
charmcraft publish-lib charms.magma_orc8r_libs.v0.orc8r_base
charmcraft publish-lib charms.magma_orc8r_libs.v0.orc8r_base_db
```

.github/workflows/orchestrator.accessd.charm.upload.yml
```diff
@@ -0,0 +1,26 @@
+name: Upload charm to Charmhub
```
gruyaume, line 1: We'd probably need a cleaner way to manage CI files that would allow for less copy/pasting, it's pretty prone to error. Probably by creating some sort of template and using `uses`. You didn't introduce this issue btw it was already present. Just a thought.

.github/workflows/orchestrator.libs.charm.upload.yml
sanchezfdezjavier: Nice catch, I assumed just publishing the charm was enough. Answering your question, I think there's no existing action to handle that, at least public. I don't know if it's worth creating it for just two libraries. Let me know what you think

Screenshot 2022-02-14 at 09 39 51

.github/workflows/orchestrator.accessd.charm.upload.yml
sanchezfdezjavier: Totally agree, it has become a mess of files. Although I don't get how to leverage `uses` for that purpose. As far as I know that "template" that you're mentioning would be another action we'll have to create additionally. 
Let me know what you mean exactly, maybe I didn't get it completely

.github/workflows/orchestrator.libs.charm.upload.yml
gruyaume: I think it'd be great in the future, but not part of this PR.

.github/workflows/orchestrator.accessd.charm.upload.yml
gruyaume: I'm not sure exactly how we'll go around and do that. But `uses`'s purpose is to refer to other actions so you don't have to copy paste stuff.

gruyaume (approved):

PR #21, opened 2022-03-04
Adds validation around pebble socket actions to catch connection errors

gruyaume:
- Adds if `can_connect` condition around pebble socket action to catch connection errors.
- Only marks status as blocked when relations are not created. 
- Marks relations as Waiting when they are created but not yet fully established.

### DIFF
--- a/orchestrator-bundle/nms-magmalte-operator/src/charm.py
+++ b/orchestrator-bundle/nms-magmalte-operator/src/charm.py
@@ -83,13 +83,19 @@ class MagmaNmsMagmalteCharm(CharmBase):
         self.unit.status = MaintenanceStatus(
             f"Configuring pebble layer for {self._service_name}..."
         )
-        plan = self._container.get_plan()
-        layer = self._pebble_layer
-        if plan.services != layer.services:
-            self._container.add_layer(self._container_name, layer, combine=True)
-            self._container.restart(self._service_name)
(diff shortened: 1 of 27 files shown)


sanchezfdezjavier (approved):

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -83,13 +83,19 @@ def _configure_pebble(self):
...
             f"Configuring pebble layer for {self._service_name}..."
         )
-        plan = self._container.get_plan()
-        layer = self._pebble_layer
-        if plan.services != layer.services:
-            self._container.add_layer(self._container_name, layer, combine=True)
-            self._container.restart(self._service_name)
-            logger.info(f"Restarted container {self._service_name}")
-            self.unit.status = ActiveStatus()
+        try:
```
jnsgruk: Could you not use `self._container.can_connect()` here? :)

Gmerold: +1
In addition to above, wouldn't it be nicer to put _can_connect()_ to **_on_magma_nms_magmalte_pebble_ready** callback and defer if container can't be reached?

gruyaume: Agreed and fixed

Gmerold (approved):

PR #28, opened 2022-04-13
Adds elasticsearch integration

gruyaume:
Orc8r libs
- All services that leverage orc8r libs will offer an interface that displays if their service is active. This new interface will be leveraged in future PR's by metricsd that requires all services to be active.

Orc8r-certifier
- Uses k8s objects instead of jinja2 templating to make calls to Kubernetes (in order to fit with the other charms)
- Moves cert management to `_on_config_changed` event handler instead of `on_install` because certs depend on `domain` config.
- Adds new validation to check that `domain` is set
- Improves unit tests
- Removes the use of the '/tmp/' directory and leverages Python dicts instead.
- Updates relation interface to `magma-orc8r-certifier`
- Removes duplicate on_install event call

Elasticsearch integration
- Adds elasticsearch integration through the use of config files (orc8r-eventd, orc8r-orchestrator)

Documentation
- Adds elasticsearch integration to how-to guides
- Simplifies deployment on AWS how-to guide to point to correct documentation instead of guiding how to deployed charmed-kubernetes.
- Minor improvements

Improvements and bug fixes:
- Removes explicit node ports in nms-nginx-proxy and orc8r-nginx
- Uses correct target port in orc8r-subscriberdb-cache
- Leverages `SIMULATE_CAN_CONNECT` in unit tests


orchestrator-bundle/documentation/how-to-guides/integrate_with_elasticsearch.md
```diff
@@ -0,0 +1,36 @@
...
+
+As of today, elasticsearch is not deployed using `juju deploy magma-orc8r`. Here we go through 
+the process of deploying elasticsearch using Helm and relating orchestrator to it.
+
+> NOTE: Relating orchestrator to elasticsearch will be modelled through a relationship once there 
+> is a charm for elasticsearch running properly on public cloud environments.
+
+### Pre-requisites
+
+- Ubuntu 20.04 machine with internet access
```
Gmerold: Could it be just any machine with internet access?

orchestrator-bundle/documentation/how-to-guides/integrate_with_elasticsearch.md
```diff
@@ -0,0 +1,36 @@
...
+- A running Magma Orchestrator instance
+- A Kubernetes cluster running on public cloud
+
+
+## Deploy Elasticsearch
+
+You can use any publicly accessible elasticsearch cluster. Here we create our own using
+helm. We deploy it in a public cloud environment to simplify the process of making it publicly
+accessible.
```
Gmerold: How about moving this part to the introduction section above and leaving only actual instructions here?

orchestrator-bundle/documentation/how-to-guides/integrate_with_elasticsearch.md
```diff
@@ -0,0 +1,36 @@
+# How-to: Integrate Magma Orchestrator with elasticsearch
+
+As of today, elasticsearch is not deployed using `juju deploy magma-orc8r`. Here we go through
```
Gmerold: How about moving the first sentence to the **NOTE**? I'm definitely not an expert with documentation, but it sounds a bit explanatory to me.
I was thinking about something like:
```
The purpose of this document is to detail the process of deploying Elasticsearch using Helm and relating Magma's Orchestrator to it. You can use any publicly accessible Elasticsearch cluster. Here we create our own using
helm. We deploy it in a public cloud environment to simplify the process of making it publicly
accessible.

**NOTE**: Once there is a Juju charm for Elasticsearch running properly on public cloud environments, Elasticsearch will be deployed and related to Orchestrator by the magma-orc8r Juju bundle.
```

orchestrator-bundle/update_libs.sh
```diff
@@ -48,6 +47,7 @@ nms-magmalte
 nms-nginx-proxy
 orc8r-bootstrapper
 orc8r-certifier
+orc8r-metricsd
 orc8r-nginx
 orc8r-orchestrator
 "
```
Gmerold: Do we need this list? I can't find it used anywhere in the code.

orchestrator-bundle/orc8r-obsidian-operator/tests/unit/test_charm.py
```diff
@@ -4,5 +4,4 @@
 
 
 def test_placeholder():
-    # TODO: write tests
```
Gmerold, line 7: How about putting here a word of explanation that the tests being done on the lib level?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         pebble_ready_event = getattr(
             self.charm.on, f"{self._service_name.replace('-', '_')}_pebble_ready"
         )
-        self._container = self.charm.unit.get_container(self._container_name)
+        self.container = self.charm.unit.get_container(self._container_name)
```
Gmerold: Should we extract libs to a separate project? 
It seems a little bit problematic to have them here if you want to update both lib and charms using it at the same time (like in this case)..

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -81,7 +81,7 @@ def __init__(
         self.charm = charm
         self.startup_command = startup_command
         self._container_name = self._service_name = self.charm.meta.name
-        self._container = self.charm.unit.get_container(self._container_name)
+        self.container = self.charm.unit.get_container(self._container_name)
```
Gmerold: As above.

orchestrator-bundle/bundle-local.yaml
```diff
@@ -236,6 +238,8 @@ relations:
   - postgresql-k8s:db
 - - nms-nginx-proxy:magmalte
   - nms-magmalte:magmalte
+- - orc8r-service-registry
+  - orc8r-metricsd
```
Gmerold: Minor thing, but personally, I find files like this one easier to work with if stuff inside it is ordered alphabetically.

orchestrator-bundle/orc8r-bundle/bundle.yaml
```diff
@@ -210,6 +212,8 @@ relations:
   - postgresql-k8s:db
 - - nms-nginx-proxy:magmalte
   - nms-magmalte:magmalte
+- - orc8r-service-registry
+  - orc8r-metricsd
```
Gmerold: As above - perhaps it would be helpful to keep things organized alphabetically.

orchestrator-bundle/orc8r-nginx-operator/src/charm.py
```diff
@@ -41,14 +41,14 @@ def __init__(self, *args):
...
             charm=self,
             ports=[
-                ("health", 80, 80, 32035),
-                ("clientcert", 8443, 8443, 30130),
-                ("open", 8444, 8444, 31694),
-                ("api", 443, 9443, 30794),
+                ("health", 80),
+                ("clientcert", 8443),
+                ("open", 8444),
+                ("api", 443, 9443),
```
Gmerold, line 47: From the docstring in `kubernetes_service_patch.py` we can read:
```
ports: a list of tuples of the form (name, port) or (name, port, targetPort)
      or (name, port, targetPort, nodePort) for every service port. If the 'targetPort'
      is omitted, it is assumed to be equal to 'port', with the exception of NodePort
      and LoadBalancer services, where all port numbers have to be specified.
```
Perhaps it would make sense to add a validator in `KubernetesServicePatch`?

orchestrator-bundle/nms-nginx-proxy-operator/src/charm.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, *args):
         self.framework.observe(self.on.remove, self._on_remove)
         self.service_patcher = KubernetesServicePatch(
             charm=self,
-            ports=[("https", 443, 443, 30760)],
+            ports=[("https", 443)],
```
Gmerold, line 41: From the docstring in `kubernetes_service_patch.py` we can read:
```
ports: a list of tuples of the form (name, port) or (name, port, targetPort)
      or (name, port, targetPort, nodePort) for every service port. If the 'targetPort'
      is omitted, it is assumed to be equal to 'port', with the exception of NodePort
      and LoadBalancer services, where all port numbers have to be specified.
```
Perhaps it would make sense to add a validator in `KubernetesServicePatch`?

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -266,6 +297,19 @@ def _pebble_layer(self) -> Layer:
...
+        elasticsearch_url_split = elasticsearch_url.split(":")
+        return elasticsearch_url_split[0], elasticsearch_url_split[1]
+
+    @property
+    def _elasticsearch_config_is_valid(self) -> bool:
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        elasticsearch_url_split = elasticsearch_url.split(":")
+        if len(elasticsearch_url_split) != 2:
+            return False
+        return True
```
Gmerold:
```suggestion
        return len(elasticsearch_url_split) != 2
```

Gmerold: Or perhaps a better option would be to test `elasticsearch_url` against a regex?
This one should do the trick: `^[a-z0-9]+:[1-9]+$`

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -98,15 +109,35 @@ def _write_config_file(self):
...
-            f'"elasticPort": {self.ELASTICSEARCH_PORT}\n'
-        )
-
-        self._container.push(f"{self.BASE_CONFIG_PATH}/orchestrator.yml", orchestrator_config)
-        self._container.push(f"{self.BASE_CONFIG_PATH}/metricsd.yml", metricsd_config)
         self._container.push(f"{self.BASE_CONFIG_PATH}/analytics.yml", analytics_config)
-        self._container.push(f"{self.BASE_CONFIG_PATH}/elastic.yml", elastic_config)
+
+    def _write_elastic_config(self):
+        if self._elasticsearch_config_is_valid:
```
Gmerold: How about moving this `if` to `_write_config_files` and only leaving the logic actually writing the config here?
If you decide to leave the `if` here, then it should be removed from `_on_elasticsearch_url_config_changed` (see the comment below).

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -98,15 +109,35 @@ def _write_config_file(self):
...
+            )
+            self._container.push(f"{self.BASE_CONFIG_PATH}/elastic.yml", elastic_config)
+        else:
+            self.unit.status = BlockedStatus(
+                "Config for elasticsearch is not valid. Format should be <hostname>:<port>"
+            )
+
+    def _on_elasticsearch_url_config_changed(self, event):
+        # TODO: Elasticsearch url should be passed through a relationship (not a config)
+        if self._elasticsearch_config_is_valid:
```
Gmerold, line 130: This `if` is redundant with the one from `_write_elastic_config`.
My vote would be on keeping this one.

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
```diff
@@ -35,7 +38,53 @@ def __init__(self, *args):
             "-logtostderr=true "
             "-v=0"
         )
-        self._orc8r_base = Orc8rBase(self, startup_command=startup_command)
+        self.orc8r_base = Orc8rBase(self, startup_command=startup_command)
```
Gmerold: This one doesn't have to be public, does it?

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
```diff
@@ -35,7 +38,53 @@ def __init__(self, *args):
...
+        elasticsearch_url_split = elasticsearch_url.split(":")
+        return elasticsearch_url_split[0], elasticsearch_url_split[1]
+
+    @property
+    def _elasticsearch_config_is_valid(self) -> bool:
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        elasticsearch_url_split = elasticsearch_url.split(":")
+        if len(elasticsearch_url_split) != 2:
+            return False
+        return True
```
Gmerold: Same as below - perhaps validating against a regex would work better here?

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
```diff
@@ -43,34 +44,110 @@ def __init__(self, *args):
...
+    def _on_pebble_ready(self, event):
+        if self._service_registry_relation_created:
+            self._configure_orc8r(event)
+        else:
+            self.unit.status = BlockedStatus("Waiting for service registry relation to be created")
+            event.defer()
+            return
+
+    def _on_relation_broken(self, event):
+        logger.info("Relation with service-registry broken - Stopping service")
```
Gmerold: Shouldn't we use something "stronger" than just an _info_? I think that's important enough to make it at least a _warning_.

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
```diff
@@ -43,34 +44,110 @@ def __init__(self, *args):
...
         return self.model.name
 
+    @property
+    def _service_registry_relation_created(self) -> bool:
+        """Checks whether charm is related to magma-service-registry."""
+        aa = self.model.get_relation("magma-service-registry")
+        logger.info(aa)
+        if not self.model.get_relation("magma-service-registry"):
+            return False
+        return True
```
Gmerold:
```suggestion
        return self.model.get_relation("magma-service-registry")
```

Gmerold: Debugging leftover?

orchestrator-bundle/orc8r-metricsd-operator/tests/unit/test_charm.py
```diff
@@ -34,9 +35,7 @@ def setUp(self):
```
Gmerold, line 39: `self.maxDiff` can be removed I guess.

orchestrator-bundle/update_libs.sh
gruyaume: It's not used in the code, it's used by you if you want to update the libs for all the charms `charmcraft update-lib`. Without it, you need to `cd` in each directory and run the `charmcraft update-lib` command and **know** which charm uses which lib.

Gmerold: Yes, but this particular list only contains charms which do not use any base lib. If we have lists for charms using base lib and base DB lib, then it seems kinda obvious that charms not listed on either of above simply do not need updating any base lib. Am I missing something?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: We could do this indeed but it'll always require 2 PR's (either 2 in the same project or 1 in each project).

orchestrator-bundle/orc8r-nginx-operator/src/charm.py
gruyaume: Here I did 2 things:
- I removed the target ports because they were the same as the port (superfluous considering the fact that kubernetes_service_patch will automatically do that if we don't specify a target port)
- I removed the nodeports because we don't need them

I don't understand you want to add a validator that does what?

Gmerold: My point is that the docstring says you **have to** specify all port numbers for _NodePort_ and _LoadBalancer_ services. If that's not the case anymore, then the docstring needs to be updated.
As for the validator - it would check whether all ports are actually specified for _NodePort_ and _LoadBalancer_ services (unless it's not needed anymore ^).

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-metricsd-operator/tests/unit/test_charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/documentation/how-to-guides/integrate_with_elasticsearch.md
gruyaume: I simplified the guide to really focus on the integration. I removed the part about deploying elasticsearch since it's our of scope of magma.

orchestrator-bundle/documentation/how-to-guides/integrate_with_elasticsearch.md
gruyaume: See other comment about this guide.

orchestrator-bundle/documentation/how-to-guides/integrate_with_elasticsearch.md
gruyaume: See my other comment about this guide.

orchestrator-bundle/update_libs.sh
gruyaume: Ahhh I get what you're saying now.

gruyaume: Fixed

orchestrator-bundle/orc8r-obsidian-operator/tests/unit/test_charm.py
gruyaume: Fixed

orchestrator-bundle/bundle-local.yaml
gruyaume: Fixed

orchestrator-bundle/orc8r-bundle/bundle.yaml
gruyaume: Fixed

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
gruyaume: Good catch, fixed

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
gruyaume: Good catch, fixed

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -102,10 +115,11 @@ def _configure_orc8r(self, event):
             pebble_layer = self._pebble_layer()
             plan = self._container.get_plan()
             if plan.services != pebble_layer.services:
-                self._container.add_layer(self._container_name, pebble_layer, combine=True)
-                self._container.restart(self._service_name)
-                logger.info(f"Restarted container {self._service_name}")
-                self.charm.unit.status = ActiveStatus()
+                self._container.add_layer(self.container_name, pebble_layer, combine=True)
+            self._container.restart(self.service_name)
```
Gmerold: Do we need to restart the container even if there are no changes done to it?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -121,15 +141,16 @@ def _configure_orc8r(self, event):
...
             pebble_layer = self._pebble_layer()
-            plan = self._container.get_plan()
+            plan = self.container.get_plan()
             if plan.services != pebble_layer.services:
-                self._container.add_layer(self._container_name, pebble_layer, combine=True)
-                self._container.restart(self._service_name)
-                logger.info(f"Restarted container {self._service_name}")
-                self.charm.unit.status = ActiveStatus()
+                self.container.add_layer(self.container_name, pebble_layer, combine=True)
+            self.container.restart(self.service_name)
```
Gmerold: As above - do we need this restart?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -201,11 +222,44 @@ def _environment_variables(self):
...
         return self.charm.model.name
+
+    def _update_relations(self):
+        if not self.charm.unit.is_leader():
+            return
+        relations = self.charm.model.relations[self.charm.meta.name]
+        for relation in relations:
+            self._update_relation_active_status(
+                relation=relation, is_active=self._service_is_running
+            )
```
Gmerold, line 240:
```suggestion
        if self.charm.unit.is_leader():
            relations = self.charm.model.relations[self.charm.meta.name]
            for relation in relations:
                self._update_relation_active_status(
                    relation=relation, is_active=self._service_is_running
                )
```

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -201,11 +222,44 @@ def _environment_variables(self):
...
+
+    def _on_relation_joined(self, event):
+        if not self.charm.unit.is_leader():
+            return
+        self._update_relation_active_status(
+            relation=event.relation, is_active=self._service_is_running
+        )
```
Gmerold, line 247:
```suggestion
    def _on_relation_joined(self, event):
        if self.charm.unit.is_leader():
            self._update_relation_active_status(
                relation=event.relation, is_active=self._service_is_running
            )
```

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -132,14 +146,47 @@ def _pebble_layer(self) -> Layer:
...
         return self.charm.model.name
+
+    def _update_relations(self):
+        if not self.charm.unit.is_leader():
+            return
+        relations = self.charm.model.relations[self.charm.meta.name]
+        for relation in relations:
+            self._update_relation_active_status(
+                relation=relation, is_active=self._service_is_running
+            )
```
Gmerold, line 168:
```suggestion
        if not self.charm.unit.is_leader():
            relations = self.charm.model.relations[self.charm.meta.name]
            for relation in relations:
                self._update_relation_active_status(
                    relation=relation, is_active=self._service_is_running
                )
```

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -132,14 +146,47 @@ def _pebble_layer(self) -> Layer:
...
+
+    def _on_relation_joined(self, event):
+        if not self.charm.unit.is_leader():
+            return
+        self._update_relation_active_status(
+            relation=event.relation, is_active=self._service_is_running
+        )
```
Gmerold, line 175:
```suggestion
    def _on_relation_joined(self, event):
        if self.charm.unit.is_leader():
            self._update_relation_active_status(
                relation=event.relation, is_active=self._service_is_running
            )
```

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -80,10 +95,14 @@ def __init__(
         super().__init__(charm, "orc8r-base")
         self.charm = charm
         self.startup_command = startup_command
-        self._container_name = self._service_name = self.charm.meta.name
-        self._container = self.charm.unit.get_container(self._container_name)
+        self.container_name = self.service_name = self.charm.meta.name
```
Gmerold, line 98: What's the reason for changing that from protected to public? I couldn't find any usage of it outside this class.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -92,10 +111,11 @@ def __init__(
         else:
             self.additional_environment_variables = {}
 
-        self._db = pgsql.PostgreSQLClient(self.charm, "db")
+        self.db = pgsql.PostgreSQLClient(self.charm, "db")
```
Gmerold, line 114: What's the reason for changing that from protected to public? I couldn't find any usage of it outside this class.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -201,11 +222,44 @@ def _environment_variables(self):
...
             "SQL_DIALECT": "psql",
-            "SERVICE_HOSTNAME": self._container_name,
+            "SERVICE_HOSTNAME": self.container_name,
         }
         environment_variables.update(sql_environment_variables)
         return environment_variables
 
     @property
-    def _namespace(self) -> str:
+    def namespace(self) -> str:
```
Gmerold, line 230: What's the reason for changing that from protected to public?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -132,14 +146,47 @@ def _pebble_layer(self) -> Layer:
...
-            "SERVICE_REGISTRY_NAMESPACE": self._namespace,
+            "SERVICE_REGISTRY_NAMESPACE": self.namespace,
         }
         environment_variables.update(self.additional_environment_variables)
         environment_variables.update(default_environment_variables)
         return environment_variables
 
     @property
-    def _namespace(self) -> str:
+    def namespace(self) -> str:
```
Gmerold, line 158: What's the reason for changing that from protected to public?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -78,12 +86,17 @@ def __init__(
         super().__init__(charm, "orc8r-base")
         self.charm = charm
         self.startup_command = startup_command
-        self._container_name = self._service_name = self.charm.meta.name
+        self.container_name = self.service_name = self.charm.meta.name
```
Gmerold, line 89: Does the _container_name_ need to be public?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -78,12 +86,17 @@ def __init__(
...
+        service_name_with_underscores = self.service_name.replace("-", "_")
         pebble_ready_event = getattr(
-            self.charm.on, f"{self._service_name.replace('-', '_')}_pebble_ready"
+            self.charm.on, f"{service_name_with_underscores}_pebble_ready"
         )
-        self._container = self.charm.unit.get_container(self._container_name)
+        relation_joined_event = getattr(
+            self.charm.on, f"{service_name_with_underscores}_relation_joined"
+        )
+        self._container = self.charm.unit.get_container(self.container_name)
```
Gmerold: I think _container_ should be public because it's used by _eventd_ and _metricsd_.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -42,14 +41,30 @@ def __init__(self, *args):
...
 ```
+
+Charms that leverage this library also need to specify a `provides` relation in their
+`metadata.yaml` file. For example:
+
+```yaml
+provides:
+  magma-orc8r-directoryd:
+    interface: magma-orc8r-directoryd
+```
```
Gmerold, line 52: Same question as above.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -44,14 +43,23 @@ def __init__(self, *args):
...
 ```
 
+Charms that leverage this library also need to specify a `provides` relation in their
+`metadata.yaml` file. For example:
+
+```yaml
+provides:
+  magma-orc8r-ha:
+    interface: magma-orc8r-ha
+```
```
Gmerold, line 53: You mean even if they don't really need it?

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_charm/metadata.yaml
```diff
@@ -18,3 +18,8 @@ resources:
     type: oci-image
     description: OCI image for magma-orc8r-ha (docker.artifactory.magmacore.org/controller:1.6.0)
     upstream-source: docker.artifactory.magmacore.org/controller:1.6.0
+
+provides:
+  magma-orc8r-ha:
+    interface:
+      magma-orc8r-ha
```
Gmerold: Not really related to this part, but what do you think about having a dummy charm for tests instead of a copy of an actual charm?

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db_charm/metadata.yaml
```diff
@@ -21,3 +21,8 @@ resources:
 requires:
   db:
     interface: pgsql
+
+provides:
+  magma-orc8r-directoryd:
+    interface:
+      magma-orc8r-directoryd
```
Gmerold: Not really related to this part, but what do you think about having a dummy charm for tests instead of a copy of an actual charm?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
Gmerold: That's true. I don't know why, but writing this comment, I was under the impression, that separation would allow us to create both PRs at the same, which is wrong...

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: How is this better?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: Yes, the idea is that services that some services rely on others (ex. metricsd relies on service-registry). When I relate metricsd to service-registry I want the metricsd service to get restarted.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
Gmerold: Just shorter.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: What do you mean?

The goal here is for all charms that leverage this library to offer an interface. I implemented this specifically because metricsd needs all other services up and running. In a next PR, I will make changes to metricsd to leverage this relation

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_charm/metadata.yaml
gruyaume: Well this is that, you can replace `ha` in your mind with `dummy` and it's a dummy charm

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: There is no reason for these to be private. The fact that some of these variables were private made it "messy" to access them outside of the library. There are situations where it would be useful for a charm to add on top of what is already in the library and leverage those variables. For example if there's an event that would trigger a change in the pebble layer just for 1 service. In this case we don't want to update the library but the charm in question needs to implement a method that will use `self.service_name` and not re-instantiate it.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: Again, there is no reason for these to be private. It made sense when they were in a charm but not here in the charm library where you could imagine a situation where a charm needs to leverage this db variable.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: Again, there is no reason for `namespace` to be private. It made sense when they were in a charm but not here in the charm library where you could imagine a situation where a charm needs to leverage this `namespace` property.

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_charm/metadata.yaml
gruyaume: I just did a search/replace in the unit tests for the libs to do that replacement, hope it makes it clearer.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: I don't know, I like the fact that it's explicit about returning when there is no leader. If you don't mind, I'll keep it as is.

Gmerold: I sure don't mind. Please treat my comments as suggestions rather than requests.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: Agree, fixed

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
Gmerold: Got it. 
That's the problem we've touched upon in the other convo - intentions are not always clear when changes need to be done to both lib and a charm.

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_charm/metadata.yaml
Gmerold: You didn't have to do that, but thanks.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
gruyaume: Yes, same answer as above

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
Gmerold: Understood. 
That's perfectly valid reason from my pov. I'm only asking this, because in one of my own PRs, I was asked not to do things 'for later'.

orchestrator-bundle/orc8r-nginx-operator/src/charm.py
gruyaume: Ok understood, in this case we'll need to change the kubernetes_service_patch doc because you don't need to specify all the ports

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -109,7 +129,7 @@ def _on_magma_orc8r_pebble_ready(self, event):
             self.charm.unit.status = BlockedStatus("Waiting for database relation to be created")
             event.defer()
             return
-        if not self._db_relation_established:
+        if not self._db_relation_created:
```
Gmerold: Now there are 2 _ifs_ with the same condition but doing two different things.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
Gmerold: Got it, but this code is run on _pebble_ready_ event and only depends on DB relation availability. Shouldn't this restart be placed in a _relation_joined_ or _relation_changed_ callback instead?

orchestrator-bundle/bundle-local.yaml
```diff
@@ -191,11 +191,17 @@ applications:
     channel: edge
     scale: 1
     trust: true
+    options:
+      metrics_count_limit: 500000
```
sanchezfdezjavier, line 199: Where does this limit value come from?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: Yeah I moved it back to what it was

orchestrator-bundle/bundle-local.yaml
gruyaume: It's the same value as in the aws deployment for this charm

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
gruyaume: That's a mistake, fixed

orchestrator-bundle/orc8r-eventd-operator/config.yaml
```diff
@@ -0,0 +1,6 @@
+options:
+  elasticsearch-url:
+    type: string
+    default:
```
sanchezfdezjavier, line 4: Is this empty `default` correct?

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
```diff
@@ -36,6 +44,47 @@ def __init__(self, *args):
...
+        if self._elasticsearch_config_is_valid:
+            if self._orc8r_base._container.can_connect():
+                self._write_config_file()
+                try:
+                    logger.info("Restarting service")
+                    self._orc8r_base._container.restart(self._orc8r_base._service_name)
+                    self.unit.status = ActiveStatus()
+                except APIError:
+                    logger.info("Service is not yet started, doing nothing")
+                    pass
```
sanchezfdezjavier: What is this `pass` doing exactly?

orchestrator-bundle/orc8r-orchestrator-operator/config.yaml
```diff
@@ -0,0 +1,6 @@
+options:
+  elasticsearch-url:
+    type: string
+    default:
```
sanchezfdezjavier, line 4: Same here

orchestrator-bundle/orc8r-eventd-operator/config.yaml
gruyaume: Yes

orchestrator-bundle/orc8r-orchestrator-operator/config.yaml
gruyaume: Yes

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
gruyaume: Yeah it's not needed, just removed it

sanchezfdezjavier (approved):

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -420,6 +405,35 @@ def _get_db_connection_string(self):
         except (AttributeError, KeyError):
             return None
 
+    @property
+    def _domain_config_is_valid(self) -> bool:
+        elasticsearch_url = self.model.config.get("domain")
+        if not elasticsearch_url:
```
Gmerold: I guess it should be something like _domain_ instead of _elasticsearch_url_ here.

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
```diff
@@ -36,6 +44,46 @@ def __init__(self, *args):
...
+    def _get_elasticsearch_config(self) -> tuple:
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        elasticsearch_url_split = elasticsearch_url.split(":")
+        return elasticsearch_url_split[0], elasticsearch_url_split[1]
+
+    @property
+    def _elasticsearch_config_is_valid(self) -> bool:
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        if not elasticsearch_url:
+            return False
```
Gmerold, line 82: Do we need this check if we have `re.match` below? I mean if _elasticsearch-url_ comes empty, `re,match` will return False anyway.

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -266,6 +298,21 @@ def _pebble_layer(self) -> Layer:
...
+    def _get_elasticsearch_config(self) -> tuple:
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        elasticsearch_url_split = elasticsearch_url.split(":")
+        return elasticsearch_url_split[0], elasticsearch_url_split[1]
+
+    @property
+    def _elasticsearch_config_is_valid(self) -> bool:
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        if not elasticsearch_url:
+            return False
```
Gmerold, line 310: Same as above.

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
gruyaume: if elasticsearch_url is None, re.match throws an error

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
gruyaume: if elasticsearch_url is None, re.match throws an error

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Agreed, fixed

orchestrator-bundle/orc8r-eventd-operator/src/charm.py
Gmerold: My bad.

Gmerold (approved):

PR #30, opened 2022-04-28
Fix: Adds tag prefix to every charm

sanchezfdezjavier:
This PR addresses the issue of the upload actions failing because of tags being already created with the same name. If two charms had the same revision number, the revision name would be also equal, ending up with duplicated names.
By adding the tag we ensure names like the following:
- `<charm name>-rev-<revision number>`

For instance, if we are uploading `eventd`'s revision number 23 we would end up with the following tag:
- `magma-orc8r-eventd-rev-23`

```yaml
      - name: Upload charm to charmhub
        uses: canonical/charming-actions/upload-charm@1.0.2
        with:
          credentials: "${{ secrets.CHARMCRAFT_AUTH }}"
          github-token: "${{ secrets.GITHUB_TOKEN }}"
          upload-image: "false"
          channel: "${{ steps.channel.outputs.name }}"
          charm-path: ./orchestrator-bundle/orc8r-eventd-operator
          tag-prefix: "magma-orc8r-eventd" <---- NEW TAG
```

**NOTE:** When talking about "revisions" I mean Charmhub revisions.

### DIFF
--- a/.github/workflows/orchestrator.accessd.charm.upload.yml
+++ b/.github/workflows/orchestrator.accessd.charm.upload.yml
@@ -33,3 +33,4 @@ jobs:
           upload-image: "false"
           channel: "${{ steps.channel.outputs.name }}"
           charm-path: ./orchestrator-bundle/orc8r-accessd-operator
+          tag-prefix: "magma-orc8r-accessd"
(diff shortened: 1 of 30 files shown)


gruyaume (approved):

Gmerold (approved):

gruyaume:
@sanchezfdezjavier This PR has been opened for 20 days. Should we merge it or close it?

Gmerold (approved):

PR #31, opened 2022-05-05
Integration tests setup

sanchezfdezjavier:
- Adds proxy-related environment variables to `tox.ini` file for every charm.
- Adds dependencies installation in `Setup operator environment` step.
- Now charms are packed using destructive mode.
- Sets `runs-on: prodstack-github-runner` to attach the integration tests on the corresponding VM in prodstack.
- Adds `ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)` just before adding relations. This prevents issues trying to relate before the charm is even deployed.
- Sets `asyncio-mode` to `auto`, to remove warnings in every integration test run. As suggested in this [issue](https://github.com/charmed-kubernetes/pytest-operator/issues/50).
- Adds missing relations in `orc8r-orchestrator` integration test.

NOTES: 
- I created another [issue](https://github.com/charmed-kubernetes/pytest-operator/issues/75) to remove the warning that was flooding the integration tests logs.
- I also created a simple [how-to-guide](https://drive.google.com/file/d/1WxpcG9HxLlj0fnmNMlzgotH-BTzed7jg/view?usp=sharing) to spin up a new Github runner in prodstack.

[Successful run](https://github.com/canonical/charmed-magma/actions/runs/2307205315)

### DIFF
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -108,7 +108,6 @@ jobs:
           #charms="$(find ./orchestrator-bundle/ -name tox.ini | grep operator)"
           #charms="./orchestrator-bundle/orc8r-bootstrapper-operator ./orchestrator-bundle/orc8r-nginx-operator"
           charms="accessd analytics configurator ctraced device directoryd dispatcher eventd ha lte metricsd obsidian smsd state streamer subscriberdb subscriberdb-cache tenants"
-
           for charm in ${charms}; do
             tox -c ./orchestrator-bundle/magma-${charm}-operator -e integration
           done
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -109,8 +109,7 @@ jobs:
         run: |
           set -eux
           #charms="$(find ./orchestrator-bundle/ -name tox.ini | grep operator)"
-          #charms="accessd analytics configurator ctraced device directoryd dispatcher eventd ha lte metricsd obsidian smsd state streamer subscriberdb subscriberdb-cache tenants"
-          charms="lte"
+          charms="accessd analytics configurator ctraced device directoryd dispatcher eventd ha lte metricsd obsidian smsd state streamer subscriberdb subscriberdb-cache tenants"
           for charm in ${charms}; do
             tox -c ./orchestrator-bundle/orc8r-${charm}-operator -e integration
           done
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -105,7 +105,7 @@ jobs:
           /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3-venv=3.8.2-0ubuntu2 python3.8-venv=3.8.10-0ubuntu1~20.04.4
           /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install cargo=0.58.0-0ubuntu1~20.04.1 libffi-dev=3.3-4 libssh2-1=1.8.0-2.1build1 libssl-dev=1.1.1f-1ubuntu2.13 libstd-rust-1.57=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 libstd-rust-dev=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 rustc=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1
 
-      # TODO: Add certifier and certifier-related charms
+      # TODO: Add certifier and certifier-related charms.
       - name: Run integration tests.
         run: |
           set -eux
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -6,50 +6,50 @@ on:
       - fix-integration-tests
 
 jobs:
-    # orchestrator-lint-report:
-    #   name: Lint report
-    #   runs-on: ubuntu-latest
-    #   steps:
-    #     - uses: actions/checkout@v2
-    #     - name: Install tox
-    #       run: pip install tox
-    #     - name: Run tests using tox
-    #       run: |
-    #         set -eux
-    #         charms="$(find ./orchestrator-bundle/ -name tox.ini)"
-    #         for charm in ${charms}; do
-    #           tox -c "${charm}" -e lint
-    #         done
+  # orchestrator-lint-report:
+  #   name: Lint report
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - name: Install tox
+  #       run: pip install tox
+  #     - name: Run tests using tox
+  #       run: |
+  #         set -eux
+  #         charms="$(find ./orchestrator-bundle/ -name tox.ini)"
+  #         for charm in ${charms}; do
+  #           tox -c "${charm}" -e lint
+  #         done
 
-    # orchestrator-static-analysis:
-    #   name: Static analysis
-    #   runs-on: ubuntu-latest
-    #   steps:
-    #     - uses: actions/checkout@v2
-    #     - name: Install tox
-    #       run: pip install tox
-    #     - name: Run tests using tox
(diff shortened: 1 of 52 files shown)


jnsgruk:
Passing comment, but there are a lot of `time.sleep` calls in here which makes me pretty uncomfortable. In a lot of cases it's straight after deploy, where you'd be better off waiting for `active/idle` in the charm? Perhaps like [this](https://github.com/canonical/alertmanager-k8s-operator/blob/730d0ee1e3308cfd1addf5575e59a55ebabf336d/tests/integration/test_upgrade_charm.py#L35)?

sanchezfdezjavier:
> Passing comment, but there is a lot of `time.sleep` calls in here which makes me pretty uncomfortable. In a lot of cases, it's straight after deploy, where you'd be better off waiting for `active/idle` in the charm? Perhaps like [this](https://github.com/canonical/alertmanager-k8s-operator/blob/730d0ee1e3308cfd1addf5575e59a55ebabf336d/tests/integration/test_upgrade_charm.py#L35)?

Agreed, I'm already working on removing the `time.sleep` calls. 

I've tried the following but didn't work out.
```python
await ops_test.model.block_until(
        lambda: APPLICATION_NAME in ops_test.model.applications, timeout=60
    )
```

Thanks for the suggestion. ~~Correct me if I'm wrong, but I'm afraid that waiting for the unit to go to `active/idle` is not an option here, as the majority of the charms that need the relation with `postgres` will go to `blocked` status before relating.~~

I think that worked, at least for some charms. Thanks! :)

orchestrator-bundle/nms-magmalte-operator/tox.ini
```diff
@@ -12,10 +12,16 @@ integration_test_path = {toxinidir}/tests/integration/
...
+deps = 
+    pytest
+    pytest-operator
 setenv =
     PYTHONPATH = {toxinidir}:{toxinidir}/lib:{[vars]src_path}
     PYTHONBREAKPOINT=ipdb.set_trace
 passenv =
+    HTTP_PROXY
+    HTTPS_PROXY
+    NO_PROXY
```
Gmerold, line 24: Did you test this setup in an env which doesn't use proxy?

.github/workflows/orchestrator.main.yml
```diff
@@ -51,20 +51,69 @@ jobs:
...
+          /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3-venv=3.8.2-0ubuntu2 python3.8-venv=3.8.10-0ubuntu1~20.04.4
+          /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install cargo=0.58.0-0ubuntu1~20.04.1 libffi-dev=3.3-4 libssh2-1=1.8.0-2.1build1 libssl-dev=1.1.1f-1ubuntu2.13 libstd-rust-1.57=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 libstd-rust-dev=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 rustc=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1
+
+      - name: Run integration tests.
+        run: |
+          set -eux
+          charms="$(find ./orchestrator-bundle/ -name tox.ini | grep operator)"
+          for charm in ${charms}; do
+            tox -c "${charm}" -e integration
+          done
```
Gmerold: I think it would be beneficial if on pull request instead of running tests for each charm separately we'd rather deploy a whole bundle and then a test suit checking whether all charms deployed successfully. This would both save us a lot of time and gave us some confidence about the bundle itself.
Lemme know what you think.

.github/workflows/orchestrator.main.yml
```diff
@@ -51,20 +51,69 @@ jobs:
...
+          /usr/bin/sg microk8s -c "microk8s kubectl -n kube-system rollout status deployment/coredns"
+          /usr/bin/sg microk8s -c "microk8s kubectl -n kube-system rollout status deployment/hostpath-provisioner"
+          echo "# Bootstrap controller"
+          sleep 180  # Microk8s needs time to initialize properly
+          /usr/bin/sg microk8s -c "juju bootstrap --debug --verbose microk8s integration-tests --model-default test-mode=true --model-default automatically-retry-hooks=false --model-default logging-config='<root>=DEBUG'  --bootstrap-constraints=''"
+          # Install dependencies
+          /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3-venv=3.8.2-0ubuntu2 python3.8-venv=3.8.10-0ubuntu1~20.04.4
+          /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install cargo=0.58.0-0ubuntu1~20.04.1 libffi-dev=3.3-4 libssh2-1=1.8.0-2.1build1 libssl-dev=1.1.1f-1ubuntu2.13 libstd-rust-1.57=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 libstd-rust-dev=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 rustc=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1
```
Gmerold: What are these dependencies?
If we need it to run tests, shouldn't it be added to tox.ini?

sanchezfdezjavier: We need them to pack some charms using destructive mode. I don't know if that answers the question.

.github/workflows/orchestrator.main.yml
sanchezfdezjavier: Agreed

orchestrator-bundle/nms-magmalte-operator/tox.ini
sanchezfdezjavier: It shouldn't be a problem if these env variables are not defined, at least that's my reasoning. Anyway, yes, it works.

gruyaume:
Linting is failing

orchestrator-bundle/orc8r-certifier-operator/tox.ini
```diff
@@ -78,4 +84,6 @@ deps =
     git+https://github.com/charmed-kubernetes/pytest-operator.git
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore {[vars]unit_test_path} --log-cli-level=INFO -s {posargs}
+    sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3-venv=3.8.2-0ubuntu2 python3.8-venv=3.8.10-0ubuntu1~20.04.4
+    sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install cargo=0.58.0-0ubuntu1~20.04.1 libffi-dev=3.3-4 libssh2-1=1.8.0-2.1build1 libssl-dev=1.1.1f-1ubuntu2.13 libstd-rust-1.57=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 libstd-rust-dev=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 rustc=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1
```
gruyaume, line 88: Humm I don't know if we should specify the versions of apt packages here. When the CI packs the charm before publishing it to charmhub, it won't use necessarily use those versions, it'll use whatever is the latest. Ex. In one week the latest rust could be 1.58. Our published charm would contain 1.58 but our integration tests would test with 1.57.

orchestrator-bundle/orc8r-ctraced-operator/tests/integration/test_integration.py
```diff
@@ -31,6 +31,7 @@ async def test_build_and_deploy(self, ops_test, setup):
         await ops_test.model.deploy(
             charm, resources=resources, application_name=APPLICATION_NAME, trust=True
         )
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
```
gruyaume, line 34: When we validate multiple behaviours we should have multiple tests. This way we know why a test failed when it failed.

PR #32, opened 2022-05-11
Test: adding unit tests to magmalte-operator to cover action events

saltiyazan:
nms-magmalte-operator has two actions that are not covered with unit tests, new test cases are added to cover them.


orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -186,3 +186,51 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
         self.harness.container_pebble_ready(container_name="magma-nms-magmalte")
 
         self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
+
+    @patch("ops.model.Container.exec")
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
+    @patch("charm.MagmaNmsMagmalteCharm._create_nms_admin_user")
+    @patch("ops.charm.ActionEvent")
+    def test_create_nms_admin_user_action(
```
sanchezfdezjavier: I suggest following the naming convention for the test methods `test_given_x_when_y_then_z`

saltiyazan: I changed the names to follow this convention :)

gruyaume:
The description should cover what has been added. You can remove the "nms-magmalte-operator has two actions that are not covered with unit tests" and reformulate to something like "Adds unit tests to validate nms-magmalte charm behaviour on juju actions"

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -186,3 +186,53 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
         self.harness.container_pebble_ready(container_name="magma-nms-magmalte")
 
         self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
+
+    @patch("ops.model.Container.exec")
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
+    @patch("charm.MagmaNmsMagmalteCharm._create_nms_admin_user")
+    @patch("ops.charm.ActionEvent")
+    def test_given_juju_action_when_callback_is_invoked_create_nms_admin_user_is_called(
```
gruyaume: when "callback is invoked"?

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -186,3 +186,53 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+        self, action_event, _create_nms_admin_user, _, mock_exec
+    ):
+        mock_exec.return_value = MockExec()
```
gruyaume: `mock_exec` is not even called as far as I can see, mocking its return value doesn't do anything

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -186,3 +186,53 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+        self.harness.charm._create_nms_admin_user_action(action_event)
+        _create_nms_admin_user.assert_called_once()
```
gruyaume: We don't want to validate that a private method is invoked, we want to validate the beaviour. Here the behaviour is that we exec a command in the container. This is important because if we change the internal implementation of `create_nms_admin_user_action`, this test would fail we we keep it as it is right now.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -186,3 +186,53 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+
+    @patch("ops.model.Container.exec")
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
+    @patch("charm.MagmaNmsMagmalteCharm._create_nms_admin_user")
+    @patch("ops.charm.ActionEvent")
+    def test_given_juju_action_when_user_creation_fails_then_action_raises_an_error(
+        self, action_event, _create_nms_admin_user, _, mock_exec
+    ):
+        _create_nms_admin_user.return_value = Exception()
```
gruyaume: Same comment as above, we don't want `_create_nms_admin_user` to raise the error, we would want `process.wait_output` to raise the error.

gruyaume (changes requested):

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: It is correct that the name was confusing. The issue has been addressed in the latest commit.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: The issue has been addressed in the latest commit, removing this mock from all the test cases where it isn't used.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: Issue has been addressed in the latest commit to validate that the _exec_ was called on the container.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +186,55 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
         relations_ready.return_value = True
-        mock_exec.return_value = MockExec()
         get_db_connection_string.return_value = self.TEST_DB_CONNECTION_STRING
 
         self.harness.container_pebble_ready(container_name="magma-nms-magmalte")
 
         self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
+
+    @patch("ops.model.Container.exec")
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
```
gruyaume, line 191: Is this mock used at all?

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +186,55 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+    @patch("ops.charm.ActionEvent")
+    def test_given_juju_action_when_create_nms_admin_user_is_called_command_executed_on_container(
```
gruyaume: the `given` is not really correct. I'd probably use something like `given_username_email_and_password are provided_when_create_nms_admin_user_juju_action_then_pebble_command_is_executed`

gruyaume: And following this, the create_nms_admin_user_action should include username, password and we should probably validate the content of the exec command.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +186,55 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+        relations_ready.return_value = False
+        action_event = Mock()
+        self.harness.charm._on_get_admin_credentials(action_event)
+        self.assertEqual(
+            action_event.fail.call_args,
+            [("Relations aren't yet set up. Please try again in a few minutes",)],
+        )
+
+    @patch("charm.MagmaNmsMagmalteCharm._relations_ready", new_callable=PropertyMock)
+    @patch("charm.MagmaNmsMagmalteCharm._get_admin_username")
```
gruyaume: We probably don't want to patch those private methods but simply the _stored content. This way if we change the internal implementation of the class and don't use the private methods `get_amdin_username` anymore, unit tests won't fail.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +186,55 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+        self.harness.charm._create_nms_admin_user_action(action_event)
+        mock_exec.assert_called_once()
+
+    @patch("ops.model.Container.exec")
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
+    @patch("ops.charm.ActionEvent")
+    def test_given_juju_action_when_user_creation_fails_then_action_raises_an_error(
+        self, action_event, _, mock_exec
+    ):
+        mock_exec.return_value = MockExec(raise_exec_error=True)
```
Gmerold: You can use `side_effect` instead of return_value. You won't have to mock the whole Exec:
`mock_exec.side_effect = ExecError(["drop"], 1, "exec", "mock")`

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +186,55 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+        with self.assertRaises(ExecError):
+            self.harness.charm._create_nms_admin_user_action(action_event)
+
+    @patch("charm.MagmaNmsMagmalteCharm._relations_ready", new_callable=PropertyMock)
```
Gmerold: You don't really have to mock this property.
Since `setUp` starts an empty harness, no existing relations is implied.

saltiyazan: addressed in the latest commit, thanks

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: addressed in the latest commit, thanks

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +186,53 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+    def test_given_juju_action_when_create_nms_admin_user_is_called_command_executed_on_container(
+        self, action_event, _, mock_exec
+    ):
+        self.harness.charm._create_nms_admin_user_action(action_event)
+        mock_exec.assert_called_once()
+
+    @patch("ops.model.Container.exec")
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
+    @patch("ops.charm.ActionEvent")
+    def test_given_juju_action_when_user_creation_fails_then_action_raises_an_error(
```
gruyaume, line 225: Specify which juju action

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: I agree that mocking the private functions like `_get_admin_username` isn't a good way to write this test. But the `_on_get_admin_credentials` that is being tested, will either:

1. return the credentials of the admin user in the master organisation as it looks for the credentials that are saved in the StoredState, and the credentials are saved in the StoredState only when the admin user in master is created when `on_magma_nms_magmalte_pebble_ready` runs.

2. generate a password for the admin@<domain name> and return those.

In this case if we mock the content of `_stored,` wouldn't this unit test fail in the future if we change the implementation not to use `_stored` to save credentials? just like if we don't use `get_admin_username` anymore? at the end of the day `get_admin_username` only returns the content of `_stored.`

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: addressed in the latest commit.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: Name has been changed and the parameters are validated in the latest commit. thanks

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -14,11 +15,18 @@
...
+            self.raise_exec_error = True
+        else:
+            self.raise_exec_error = False
+
     def exec(self, *args, **kwargs):
         pass
 
     def wait_output(self, *args, **kwargs):
-        pass
+        if hasattr(self, "raise_exec_error") and self.raise_exec_error:
```
gruyaume: If I'm not mistaken, the hasattr will always be true here since we have an if/else statement in the emit that sets it either way.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -33,6 +41,11 @@ class TestCharm(unittest.TestCase):
         "user=test_db_user"
     )
     TEST_DOMAIN_NAME = "test.domain.com"
+    CREATE_USER_ACTION_EVENT_PARAMS = {
```
gruyaume: Does this variable need to be global? It's only used at 1 place.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -180,9 +191,68 @@ def test_given_charm_when_pebble_ready_event_emitted_and_relations_are_establish
...
+    @patch("charm.MagmaNmsMagmalteCharm._get_db_connection_string", new_callable=PropertyMock)
+    @patch("ops.charm.ActionEvent")
+    def test_given_juju_action_when_user_creation_fails_then_action_raises_an_error(
+        self, action_event, _, mock_exec
+    ):
+        mock_exec.side_effect = ExecError(["drop"], 1, "exec error", "mock exec error")
+        with self.assertRaises(ExecError):
+            self.harness.charm._create_nms_admin_user_action(action_event)
+
+    def test_given_juju_action_when_relation_is_not_realized_then_get_admin_credentials_fails(
```
gruyaume: What does "Realized" mean. Relation can be "created", "joined" or even "ready" in our code but "realized" I don't know.

saltiyazan: I think I used "realised" to have a more general term, as the `_relations_ready` in the charm code waits for relations to be "created", "established" or "ready" and it will return false if any of those is false. But I guess I can use the term "ready" just like the function does in its name.

gruyaume: Yeah I'd use ready for consistency

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: Correct, I addressed it in the latest commit.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: Addressed in the latest commit :)

gruyaume (approved):

PR #33, opened 2022-05-16
Upgrading magma core version to 1.7

saltiyazan:
Currently magma-core 1.6 is used, update the metadata.yaml file of all charms that use magma-core to version 1.7 which is the current stable version


gruyaume:
Have you tested it and it works fine, there's no new service in 1.7 that weren't there in 1.6? And were these new images uploaded to charmhub?

PR #35, opened 2022-05-26
Implements TLS relationship

gruyaume:
- Removes logic to create certificate in orc8r-certifier
- Implements tls-certificates relationship in orc8r-certifier, nms-magmalte, nms-nginx-proxy through the use of the tls_certificates_interface library


orchestrator-bundle/orc8r-certifier-operator/config.yaml
```diff
@@ -1,69 +1,9 @@
...
-    type: boolean
-    default: true
-    description: |
-      If set to true, set of self-signed SSL certs will be created.
-      For production deployments, trusted certs should be provided instead.
   passphrase:
     type: string
     default: password123
     description: |
       Passphrase used to open the admin_operator.pfx certificate
```
Gmerold, line 10: Do we still need this? 
I can't find any usage of it.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -73,6 +50,40 @@ def __init__(self, *args):
...
+        self.framework.observe(
+            self.on.certificates_relation_joined, self._on_certificates_relation_joined
+        )
+
+    def _on_certificates_relation_joined(self, event):
+        domain_name = self.model.config.get("domain")
+        if not self._domain_config_is_valid:
+            logger.info("Domain config is not valid")
+            event.defer()
+            return
```
Gmerold, line 154: How about putting the charm to blocked status instead of just deferring the event? 
This kind of error definitely requires operator's intervention.
Actually there's a good example of what I mean in lines 107-111.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -73,6 +50,40 @@ def __init__(self, *args):
...
+        self.certificates.request_certificate(
+            cert_type="server",
+            common_name=domain_name,  # type: ignore[arg-type]
+        )
+
+    def _on_certificate_available(self, event):
+        logger.info("Certificate is available")
+        if not self._container.can_connect():
+            logger.info("Cant connect to container - Won't push certificates to workload")
```
Gmerold, line 164: I think we could use some higher severity (at least warning) for logging things like that.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -412,33 +238,6 @@ def _domain_config_is_valid(self) -> bool:
             return False
```
Gmerold, line 380: We could use [validators](https://github.com/kvesteri/validators) library for stronger validation.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -91,37 +102,43 @@ def _on_install(self, event):
...
         self._configure_magma_orc8r_certifier(event)
 
+    def _on_config_changed(self, event):
+        if not self._domain_config_is_valid:
+            self.unit.status = BlockedStatus("Config 'domain' is not valid")
+            logger.warning("Config 'domain' not valid")
+            event.defer()
+            return
+
+    def _certs_are_stored(self) -> bool:
```
Gmerold: Why not a property?

orchestrator-bundle/orc8r-certifier-operator/config.yaml
gruyaume: Correct, removed. We'll need it "somewhere" but not sure where yet

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: We could place it in blocked and not defer. This would imply requesting certificates in the handling of the config changed event

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: You don't like my validations? :laughing:

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Agreed, fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: So how's the weather in Canada? :D

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: Hmmm, this sounds even better. You'd probably want new certs if you change the domain anyway...

PR #36, opened 2022-05-27
Fix: Adding relation between orc8r-orchestrator and orc8r-prometheus-cache

Gmerold:
This PR is 1/2 of the fix for not working metrics.
This PR should be merged after [Prometheus-edge-hub PR](https://github.com/canonical/prometheus-edge-hub-operator/pull/9)

- Adds relation between orc8r-orchestrator and orc8r-prometheus-cache to make sure they get deployed in correct order
- Fixes relation name fetching in orc8r_base libraries
- Adds support for checking if required relations have been satisfied prior to configuring pebble to the orc8r_base lib (will be need in the next PR) 
- Renames some of the relations provided by the charms (rest will come with the next PR)
- Other minor updates


orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -439,6 +460,42 @@ def _encode_in_base64(byte_string: bytes):
         """Encodes given byte string in Base64"""
         return base64.b64encode(byte_string).decode("utf-8")
 
+    def _on_magma_orc8r_certifier_relation_joined(self, event: RelationJoinedEvent):
```
gruyaume, line 463: Do we want to only update the service status when the relation is joined? If the status is `down` when the relation joins and then changes to `up` after a few seconds, it will never notify.

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -279,9 +286,9 @@ def _nms_certs_mounted(self) -> bool:
     def _domain_name(self):
         """Returns domain name provided by the orc8r-certifier relation."""
         try:
-            certifier_relation = self.model.get_relation("certifier")
-            units = certifier_relation.units
-            return certifier_relation.data[next(iter(units))]["domain"]
+            certifier_relation = self.model.get_relation("magma-orc8r-certifier")
```
gruyaume, line 289: We may want to add a limit to the number of relations to `oc8r-cerfifier` to 1 in metadata.yml. This should never be an issue in the context of our charms, but still. If somebody tries to relate 2 certifiers to this charm using the `magma-orc8r-certifier` relation name, this line will crash.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: True, fixed.

orchestrator-bundle/nms-magmalte-operator/src/charm.py
Gmerold: This actually applies to all dependencies in all of our charms. 
I've added limits all across the project.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -439,6 +460,45 @@ def _encode_in_base64(byte_string: bytes):
...
+        if not self.unit.is_leader():
+            return
```
sanchezfdezjavier, line 465: When would the unit not be the leader? I assume this is not taking scaling into account right?

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -190,3 +206,23 @@ def _update_relation_active_status(self, relation, is_active: bool):
...
+    @property
+    def _relations_ready(self) -> bool:
+        """Checks whether required relations are ready."""
+        missing_relations = [
+            relation for relation in self.required_relations if not self._relation_active(relation)
+        ]
+        if missing_relations:
+            msg = f"Waiting for relations: {', '.join(missing_relations)}"
+            self.charm.unit.status = BlockedStatus(msg)
+            return False
```
sanchezfdezjavier:
```suggestion
        if missing_relations := [relation for relation in self.required_relations if not self._relation_active(relation)]:
            msg = f"Waiting for relations: {', '.join(missing_relations)}"
            self.charm.unit.status = BlockedStatus(msg)
            return False
```
A bit shorter?

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -5,7 +5,7 @@
 from unittest.mock import Mock, PropertyMock, patch
 
 from ops import testing
-from ops.model import BlockedStatus
+from ops.model import WaitingStatus
```
sanchezfdezjavier: Why are we changing `BlockedStatus` to `WaitingStatus`? I assume because the "default" use case is deployment with the bundle, so no "human interaction" is required?

orchestrator-bundle/orc8r-orchestrator-operator/metadata.yaml
```diff
@@ -23,9 +23,17 @@ resources:
...
+  magma-orc8r-orchestrator:
+    interface: magma-orc8r-orchestrator
+
 requires:
-  certifier:
+  magma-orc8r-certifier:
     interface: magma-orc8r-certifier
+    limit: 1
+  metrics-endpoint:
+    interface: prometheus_scrape
```
sanchezfdezjavier, line 35: @gruyaume shouldn't we split words with either underscore or dash, but the same for every interface? Is there any reason why the edge hub is using underscore here?

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
Gmerold: No, that's simply me being wrong... No idea why did I change that. It should be blocked not waiting. 
Will fix it.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: It actually does. If you have more units of an application, there can be only one leader. 
In case of a single unit, I guess the only moment when there's no leader, is right after application deployment, before leader election happens.

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
Gmerold: Fixed.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
sanchezfdezjavier: Oh perfect, then got it. Thanks

gruyaume (approved):

orchestrator-bundle/orc8r-orchestrator-operator/metadata.yaml
gruyaume: Well some interfaces we didn't choose their names. For example prometheus_scrape was defined by the LMA team and here we are simply using it.

PR #37, opened 2022-05-30
Accessd integration tests

sanchezfdezjavier:
- Refactors integration tests for accessd.
- Adds proxy-related environment variables to `tox.ini`.
- Now wee pack using `--destructive-mode`.
- Adds `ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)` just before adding relations. This prevents issues trying to relate before the charm is even deployed.
- Sets `asyncio-mode` to `auto`, to remove warnings. As suggested in this [issue](https://github.com/charmed-kubernetes/pytest-operator/issues/50).

NOTES: 
- I created another [issue](https://github.com/charmed-kubernetes/pytest-operator/issues/75) to remove the warning that was flooding the integration tests logs.


orchestrator-bundle/orc8r-accessd-operator/tox.ini
```diff
@@ -75,7 +78,10 @@ description = Run integration tests
 deps =
     git+https://github.com/juju/python-libjuju.git
     pytest
+    pytest-operator
     git+https://github.com/charmed-kubernetes/pytest-operator.git
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore {[vars]unit_test_path} --log-cli-level=INFO -s {posargs}
+    /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3-venv=3.8.2-0ubuntu2 python3.8-venv=3.8.10-0ubuntu1~20.04.4
+    /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install cargo=0.58.0-0ubuntu1~20.04.1 libffi-dev=3.3-4 libssh2-1=1.8.0-2.1build1 libssl-dev=1.1.1f-1ubuntu2.13 libstd-rust-1.57=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 libstd-rust-dev=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1 rustc=1.57.0+dfsg1+llvm-0ubuntu1~20.04.1
```
gruyaume: Are you sure those are all needed? example rust, cargo, libssl?

sanchezfdezjavier: You're right, those are leftovers from `certifier`, removed.

orchestrator-bundle/orc8r-accessd-operator/tests/integration/test_integration.py
```diff
@@ -22,15 +22,22 @@ async def setup(self, ops_test):
...
             f"{CHARM_NAME}-image": METADATA["resources"][f"{CHARM_NAME}-image"]["upstream-source"],
         }
         await ops_test.model.deploy(
             charm, resources=resources, application_name=APPLICATION_NAME, trust=True
         )
+
+    async def test_wait_for_blocked_status(self, ops_test, build_and_deploy_charm):
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
+
+    @pytest.mark.abort_on_fail
```
Gmerold: If we're unsure about the tests order, maybe it would be better not to use this decorator on any of the tests? 
I mean if this one runs first and fails, we don't get feedback from other tests.
I'd mark `build_and_deploy_charm` with this decorator tho.
What do you think?

orchestrator-bundle/orc8r-accessd-operator/tox.ini
```diff
@@ -75,7 +78,9 @@ description = Run integration tests
 deps =
     git+https://github.com/juju/python-libjuju.git
     pytest
+    pytest-operator
     git+https://github.com/charmed-kubernetes/pytest-operator.git
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore {[vars]unit_test_path} --log-cli-level=INFO -s {posargs}
+    /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3-venv=3.8.2-0ubuntu2 python3.8-venv=3.8.10-0ubuntu1~20.04.4
```
Gmerold: Why do we need two different venvs?

sanchezfdezjavier: I'm not sure, I've just used the command that is executed when packing.

orchestrator-bundle/orc8r-accessd-operator/tests/integration/test_integration.py
sanchezfdezjavier: Agreed, fixed. It had no sense `@pytest.mark.abort_on_fail` to be there.

Gmerold (approved):

gruyaume (approved):

PR #38, opened 2022-05-31
Fixing bundle upload; adding libs upload

Gmerold:
- Fixes bundle upload
- Adds orc8r-libs publishing

### DIFF
--- a/.github/workflows/orchestrator.libs.charm.upload.yml
+++ b/.github/workflows/orchestrator.libs.charm.upload.yml
@@ -19,3 +19,15 @@ jobs:
           credentials: "${{ secrets.CHARMCRAFT_AUTH }}"
           charm-path: ./orchestrator-bundle/orc8r-libs
           upload-image: "false"
+
+      - name: Publish libs
+        env:
+          CHARMCRAFT_AUTH: "${{ secrets.CHARMCRAFT_AUTH }}"
+        run: |
+          set -eux
+          orc8r_libs_dir="./orchestrator-bundle/orc8r-libs/lib"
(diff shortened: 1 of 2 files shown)


.github/workflows/orchestrator.libs.charm.upload.yml
```diff
@@ -19,3 +19,15 @@ jobs:
...
           charm-path: ./orchestrator-bundle/orc8r-libs
           upload-image: "false"
+
+      - name: Publish libs
+        env:
+          CHARMCRAFT_AUTH: "${{ secrets.CHARMCRAFT_AUTH }}"
+        run: |
+          set -eux
+          orc8r_libs_dir="./orchestrator-bundle/orc8r-libs/lib"
+          libs="$(find ${orc8r_libs_dir}/charms/magma_orc8r_libs/v0/ -name *.py -exec sh -c 'printf "%s\n" "${0%.*}"' {} ';' | sed "s#${orc8r_libs_dir}/##g" | sed "s#/#.#g")"
```
gruyaume: Wouldn't it be simpler to read if the lib paths were just listed insted of finding them with this script?

Gmerold: It would be, but then we'd have to remember about changing the workflow every time new lib is added (if it's ever the case ;) ).
If you think hardcoded list is still better, that's fine by me.

gruyaume: Yeah I don't think we'll need to add that many more libs, I'd rather have them hardcoded

Gmerold: Done.

gruyaume (approved):

PR #39, opened 2022-06-01
Adding DB connection check before starting the service

Gmerold:
- Adds a DB connection check in addition to other DB relation checks, to make sure that the DB is actually accessible before starting Magma service

### DIFF
--- a/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
+++ b/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
@@ -56,6 +56,7 @@ provides:
 import logging
 
 import ops.lib
+import psycopg2  # type: ignore[import]
 from ops.charm import CharmBase
 from ops.framework import Object
 from ops.model import ActiveStatus, MaintenanceStatus, ModelError, WaitingStatus
@@ -70,7 +71,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 11
 
 
 logger = logging.getLogger(__name__)
@@ -184,11 +185,21 @@ class Orc8rBase(Object):
 
     @property
     def _db_relation_established(self) -> bool:
-        """Validates that database relation is ready (that there is a relation and that credentials
-        have been passed)."""
-        if not self._get_db_connection_string:
+        """Validates that database relation is ready (that there is a relation, credentials have
+        been passed and the database can be connected to)."""
+        db_connection_string = self._get_db_connection_string
+        if not db_connection_string:
             return False
-        return True
+        try:
+            psycopg2.connect(
+                f"dbname='{self.DB_NAME}' "
+                f"user='{db_connection_string.user}' "
+                f"host='{db_connection_string.host}' "
+                f"password='{db_connection_string.password}'"
+            )
+            return True
+        except psycopg2.OperationalError:
+            return False
 
     @property
     def _get_db_connection_string(self):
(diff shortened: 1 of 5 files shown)


orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -184,11 +185,21 @@ def _on_database_relation_joined(self, event):
...
-        """Validates that database relation is ready (that there is a relation and that credentials
-        have been passed)."""
-        if not self._get_db_connection_string:
+        """Validates that database relation is ready (that there is a relation, credentials have
+        been passed and the database can be connected to)."""
+        db_connection_string = self._get_db_connection_string
+        if not db_connection_string:
+            return False
+        try:
+            psycopg2.connect(
```
gruyaume, line 194: cool :)

orchestrator-bundle/orc8r-libs/test-requirements.txt
```diff
@@ -0,0 +1 @@
+psycopg2-binary
```
gruyaume: Why isn't this in the requirements file instead?

orchestrator-bundle/orc8r-libs/tox.ini
```diff
@@ -70,6 +70,7 @@ deps =
     pytest
     coverage[toml]
     -r{toxinidir}/requirements.txt
+    -r{toxinidir}/test-requirements.txt
```
gruyaume: Same, is this really needed?

orchestrator-bundle/orc8r-libs/test-requirements.txt
Gmerold: `requirements.txt` doesn't allow using binaries - if you put one there, you can't build a charm. On the other hand, this is needed for tests, so I figured I'll put it in here.
Logical question would be "why didn't I use sources to build psycopg2?" That's because then it requires additional packages (only available through apt) in runtime, so I could build the charm, but it would fail during the deploy.

orchestrator-bundle/orc8r-libs/tox.ini
Gmerold: See answer above.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
Gmerold: Thanks :)

orchestrator-bundle/orc8r-libs/test-requirements.txt
gruyaume: I just placed this lib inside of orc8r-smsd, added `psycopg2-binary` to its reuqirements and added this block to the charmcraft.yaml file:
```
charm-binary-python-packages:
      - psycopg2-binary
```
The charm packed just fine. I haven't deployed it though and tested it further.

Gmerold: Well, yes, in that configuration it will build, because charmcraft will first install packages from `charmcraft.yaml` and then from `requirements.txt` (well, kinda, because pip will detect that the dependency is already satisfied, so it won't install it again). IMO that can be confusing; for instance - if someone wants to hardcode the dependency version in `requirements.txt` that won't work.
How 'bout I put this test dependency into the `tox.ini` directly?

gruyaume: Yeah I understand now. I don't know, I still think putting it in requirements.txt is just simpler. You may want to add a comment next to it like # Package only used for testing, charm uses charm-binary-python-packages.

Gmerold: Ok, done.

gruyaume (approved):

PR #45, opened 2022-06-03
Fixing configs being written to the container before it's up

Gmerold:
Fixes error caused by the fact that we're trying to write config files to a container which is not yet running

### DIFF
--- a/orchestrator-bundle/orc8r-analytics-operator/tests/unit/test_charm.py
+++ b/orchestrator-bundle/orc8r-analytics-operator/tests/unit/test_charm.py
@@ -25,6 +25,7 @@ class Test(unittest.TestCase):
     )
     def setUp(self):
         self.harness = testing.Harness(MagmaOrc8rAnalyticsCharm)
+        self.harness.set_can_connect("magma-orc8r-analytics", True)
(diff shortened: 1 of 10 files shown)


orchestrator-bundle/orc8r-analytics-operator/tests/unit/test_charm.py
```diff
@@ -25,6 +25,7 @@ class Test(unittest.TestCase):
     )
     def setUp(self):
         self.harness = testing.Harness(MagmaOrc8rAnalyticsCharm)
+        self.harness.set_can_connect("magma-orc8r-analytics", True)
```
gruyaume, line 28: Ohh I didn't know this method was available, it'll be useful :)

gruyaume (approved):

orchestrator-bundle/orc8r-analytics-operator/tests/unit/test_charm.py
Gmerold: Yep, it is useful indeed. It needs to be used together with `testing.SIMULATE_CAN_CONNECT = True` though.

PR #48, opened 2022-06-09
Bug fix: Charms crashing before relation is created

gruyaume:
Fixes bug in library and in orchestrator where charms requiring relations would crash before the relation was created.

### DIFF
--- a/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
+++ b/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
@@ -216,7 +216,7 @@ class Orc8rBase(Object):
             relation = self.model.get_relation(relation_name)
             if relation:
                 units = relation.units
-                return bool(relation.data[next(iter(units))]["active"])
+                return relation.data[next(iter(units))]["active"] == "True"
             else:
                 return False
         except (KeyError, StopIteration):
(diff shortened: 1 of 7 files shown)


orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -218,10 +211,13 @@ def _relations_ready(self) -> bool:
...
-    def _relation_active(self, relation) -> bool:
+    def _relation_active(self, relation_name: str) -> bool:
         try:
-            rel = self.model.get_relation(relation)
-            units = rel.units  # type: ignore[union-attr]
-            return bool(rel.data[next(iter(units))]["active"])  # type: ignore[union-attr]
+            relation = self.model.get_relation(relation_name)
+            if relation:
+                units = relation.units
+                return bool(relation.data[next(iter(units))]["active"])
```
Gmerold:
```suggestion
                return relation.data[next(iter(units))]["active"] == "True"
```

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -220,11 +220,14 @@ def _relations_ready(self) -> bool:
...
-    def _relation_active(self, relation: str) -> bool:
+    def _relation_active(self, relation_name: str) -> bool:
         try:
-            rel = self.model.get_relation(relation)
-            units = rel.units  # type: ignore[union-attr]
-            return bool(rel.data[next(iter(units))]["active"])  # type: ignore[union-attr]
+            relation = self.model.get_relation(relation_name)
+            if relation:
+                units = relation.units
+                return bool(relation.data[next(iter(units))]["active"])
```
Gmerold:
```suggestion
                return relation.data[next(iter(units))]["active"] == "True"
```

Gmerold (approved):

PR #53, opened 2022-06-22
Bundle integration tests

sanchezfdezjavier:
- Adds integration tests for `orc8r-bundle`
- Modifies `orchestrator.main.yaml` workflow file for running integration tests for the whole bundle instead of individual integration tests for each charm.
- Toggles `--model-default automatically-retry-hooks` to `True` while bootstrapping the controller for solving the issue of events being lost while bundle deployment.
- For running `bundle` integration tests go under `charmed-magma/orchestrator-bundle/orc8r-bundle` and run 
```bash
tox -e integration
```

### DIFF
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -120,7 +120,7 @@ jobs:
           set -eux
           ./orchestrator-bundle/build.sh
 
-      - name: Deploy bundle and wait for idle
+      - name: Run bundle integration tests
         run: |
           tox -c "./orchestrator-bundle/orc8r-bundle/tox.ini" -e integration
 
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -120,14 +120,9 @@ jobs:
           set -eux
           ./orchestrator-bundle/build.sh
 
-      # - name: Run integration tests
-      #   run: |
-      #     set -eux
-
-      #     charms="$(find ./orchestrator-bundle/ -name tox.ini | grep operator | grep -wv orc8r-orchestrator)"
-      #     for charm in ${charms}; do
-      #       tox -c "${charm}" -e integration
-      #     done
+      - name: Deploy and wait for idle
+        run: |
+          tox -c "./orchestrator-bundle/orc8r-bundle/tox.ini" -e integration
 
       - name: Clean up
         if: always()
--- a/.github/workflows/orchestrator.main.yml
+++ b/.github/workflows/orchestrator.main.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Run tests using tox
         run: |
           set -eux
-          charms="$(find ./orchestrator-bundle/ -name tox.ini -not -path "./orchestrator-bundle/orc8r-bundle/*")"
+          charms="$(find ./orchestrator-bundle/ -name tox.ini)"
           for charm in ${charms}; do
             tox -c "${charm}" -e static
           done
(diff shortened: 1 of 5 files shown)


.github/workflows/orchestrator.main.yml
```diff
@@ -109,20 +109,27 @@ jobs:
...
+      - name: Build charms
         run: |
           set -eux
+          ./orchestrator-bundle/build.sh
 
-          charms="$(find ./orchestrator-bundle/ -name tox.ini | grep operator)"
-          for charm in ${charms}; do
-            tox -c "${charm}" -e integration
-          done
+      - name: Deploy and wait for idle
```
Gmerold:
```suggestion
      - name: Run integration tests
```

.github/workflows/orchestrator.main.yml
```diff
@@ -109,20 +109,27 @@ jobs:
...
 
           # Install dependencies
           /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install python3.8-venv=3.8.10-0ubuntu1~20.04.4
           /usr/bin/sudo apt-get --no-install-recommends -y -oDpkg::Use-Pty=0 --allow-downgrades install libffi-dev libssh2-1 libssl-dev libssl-dev libstd-rust-1.57 cargo libstd-rust-dev rustc
 
-      - name: Run integration tests
+      - name: Build charms
         run: |
           set -eux
+          ./orchestrator-bundle/build.sh
```
Gmerold, line 117: Did you try running builds in parallel? I wonder if we have enough computing power on our runners to actually benefit from that?

orchestrator-bundle/orc8r-bundle/pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[tool.coverage.run]
+branch = true
+
+[tool.coverage.report]
+show_missing = true
```
Gmerold: This can be removed as we're not checking the coverage.

.github/workflows/orchestrator.main.yml
sanchezfdezjavier: I didn't, but now we are leveraging the cache from the runner and the build times are pretty short. But I give it a try when I have more VMs in prodstack, I think it's worth it

orchestrator-bundle/orc8r-bundle/pyproject.toml
sanchezfdezjavier: True, removed.

.github/workflows/orchestrator.main.yml
sanchezfdezjavier: I change it to `Run bundle integration test`, let me know if you agree. That way is easier to spot they are bundle tests in Github's actions tab

Gmerold: Yup, that works too.

.github/workflows/orchestrator.main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Run tests using tox
         run: |
           set -eux
-          charms="$(find ./orchestrator-bundle/ -name tox.ini)"
+          charms="$(find ./orchestrator-bundle/ -name tox.ini -not -path "./orchestrator-bundle/orc8r-bundle/*")"
```
Gmerold: Is there any reason why we don't wanna run static analysis on the bundle?

sanchezfdezjavier: There's not, my bad. Fixed.

Gmerold (approved):

PR #55, opened 2022-06-24
Fix service status in relation data 2

Gmerold:
- Updates base libs in all charms to leverage service status published in the relation data
- Creates relation between orc8r-metricsd and orc8r-orchestrator to ensure proper metrics handling
- Updates bundles accordingly
- Other minor updates


orchestrator-bundle/orc8r-ctraced-operator/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -77,7 +78,7 @@ def __init__(self, *args):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 9
+LIBPATCH = 11
```
sanchezfdezjavier, line 81: Why are we jumping version 10?

sanchezfdezjavier (approved):
LGTM

PR #65, opened 2022-07-08
Consolidates documentation under the bundle's readme / charmhub page.

gruyaume:
Documentation updates:
- Documentation used to be spread through a documentation subdirectory as well as in the charmed-orc8r readme. This change consolidates documentation under the bundle's readme. This readme is also what is published on Charmhub.
- Updates "Contributing" guide to current state of the project and removes the "Roadmap" section as it was unmaintained.
- Removes duplicate License
- Fills in license content


orchestrator-bundle/orc8r-bundle/README.md
```diff
@@ -11,11 +9,38 @@ and it has been tested with all major public cloud platforms.
...
+
+This how-to guide can be used to deploy Magma's Orchestrator on any cloud environment. It contains
+steps to set up a Kubernetes cluster, bootstrap a Juju controller, deploy charmed operators for
+Magma Orchestrator and configure DNS A records. For more information on Charmed Magma, please visit
+the project's [homepage](https://github.com/canonical/charmed-magma).
+
+### Pre-requisites
+
+- Ubuntu 20.04 machine with internet access
+- A public domain
```
sanchezfdezjavier, line 22: People still try to deploy using dummy domains they don't own like `domain.com`, I'm aware it's redundant but also I think it is worth for avoiding unnecessarily questions/support
```suggestion
- A previously purchased public domain
```

sanchezfdezjavier (dismissed):
Really like the new cleaner approach

Gmerold (commented):
Now, that we have our deployment how-tos being part of official Magma documentation, wouldn't it make sense to just refer the official doc in `orchestrator-bundle/orc8r-bundle/README.md`? In case anything changes, it would only require changing one document.

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: I don't know , `a public domain` is what is needed, whether it was purchased, gotten for free or stolen

gruyaume:
> Now, that we have our deployment how-tos being part of official Magma documentation, wouldn't it make sense to just refer the official doc in `orchestrator-bundle/orc8r-bundle/README.md`? In case anything changes, it would only require changing one document.

Following our conversation I'd keep it as is. This is already an improvement (2 documents instead of 3).

Gmerold (approved):

orchestrator-bundle/orc8r-bundle/README.md
sanchezfdezjavier: Agreed, "previously purchased" it's not the best choice. Still, I remember several people finding themselves not reaching the NMS UI because they missed the public domain and A records step. Don't you think some sort of clarification won't be bad? Maybe something like
```suggestion
- A public domain. If you don't have access to one, you can purchase it from [Google Domains](https://domains.google/)
```
At least this avoids people using dummy domains like `example.com`. I don't know, I might be exaggerating the issue. Let me know what you think

Gmerold: Personally, I don't think it needs any additional explanation. In your example, `example.com` is not a public domain (unless you actually own it ;) ).

sanchezfdezjavier: You've convinced me.

sanchezfdezjavier (approved):

PR #68, opened 2022-07-15
Chore: Remove create orchestrator admin user action and create user by default

saltiyazan:
- Remove the `create-orchestrator-admin-user` action
- Creates the user automatically
- Checks if accessd is up and if the orchestrator container can be connected to before creating the user
- Adds unit tests


gruyaume:
As a general comment, this PR and [this one](https://github.com/canonical/charmed-magma/pull/66) will for sure have conflicts as both touch lots of the orc8r-orchestrator charm.

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -237,6 +233,13 @@ def _on_magma_orc8r_orchestrator_pebble_ready(self, event: PebbleReadyEvent):
             event.defer()
             return
         self._configure_orc8r(event)
+        if not self._container.can_connect():
```
gruyaume: _configure_orc8r already checks that we `can_connect`

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
```diff
@@ -126,6 +144,61 @@ def test_given_new_charm_when_on_install_event_then_analytics_config_file_is_cre
...
+        ]
+        self.assertIn(call_command, args)
+
+    @patch("ops.model.Container.exec", new_callable=Mock)
+    @patch("ops.model.Container.push")
+    @patch("charm.MagmaOrc8rOrchestratorCharm._relations_ready", PropertyMock(return_value=False))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._namespace", PropertyMock(return_value="whatever"))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._relations_created", PropertyMock(return_value=True))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._nms_certs_mounted", PropertyMock(return_value=True))
+    def test_given_relations_not_ready_on_pebble_ready_event_then_orchestrator_admin_user_is_not_created(  # noqa: E501
```
gruyaume:
```suggestion
    def test_given_relations_not_ready_when_pebble_ready_event_then_orchestrator_admin_user_is_not_created(  # noqa: E501
```

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
```diff
@@ -126,6 +144,61 @@ def test_given_new_charm_when_on_install_event_then_analytics_config_file_is_cre
...
+        self.harness.charm.on.magma_orc8r_orchestrator_pebble_ready.emit(container)
+        patch_exec.assert_not_called()
+
+    @patch("ops.model.Container.exec", new_callable=Mock)
+    @patch("ops.model.Container.push")
+    @patch("charm.MagmaOrc8rOrchestratorCharm._relations_ready", PropertyMock(return_value=True))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._namespace", PropertyMock(return_value="whatever"))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._relations_created", PropertyMock(return_value=True))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._nms_certs_mounted", PropertyMock(return_value=True))
+    def test_given_orchestrator_service_not_running_on_pebble_ready_event_then_orchestrator_admin_user_is_not_created(  # noqa: E501
```
gruyaume:
```suggestion
    def test_given_orchestrator_service_not_running_when_pebble_ready_event_then_orchestrator_admin_user_is_not_created(  # noqa: E501
```

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
```diff
@@ -126,6 +144,61 @@ def test_given_new_charm_when_on_install_event_then_analytics_config_file_is_cre
...
+    @patch("charm.MagmaOrc8rOrchestratorCharm._namespace", PropertyMock(return_value="whatever"))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._relations_created", PropertyMock(return_value=True))
+    @patch("charm.MagmaOrc8rOrchestratorCharm._nms_certs_mounted", PropertyMock(return_value=True))
+    def test_given_relations_ready_and_container_can_connect_when_on_pebble_ready_event_then_orchestrator_admin_user_is_created(  # noqa: E501
+        self, patch_push, patch_exec
+    ):
+        patch_exec.return_value = MockExec()
+        container = self.harness.model.unit.get_container("magma-orc8r-orchestrator")
+        self.harness.set_can_connect(container, True)
+        self.harness.charm.on.magma_orc8r_orchestrator_pebble_ready.emit(container)
```
gruyaume: There is a self.harnes.container_pebble_ready (or something like so) which you can use instead of emitting the event directly. This will also set your "connect" to true so you can also remove that line.

orchestrator-bundle/bundle-local.yaml
```diff
@@ -240,6 +240,8 @@ relations:
   - orc8r-alertmanager:alerting
 - - orc8r-prometheus:metrics-endpoint
   - orc8r-prometheus-cache:metrics-endpoint
+- - orc8r-orchestrator:magma-orc8r-accessd
```
gruyaume, line 250: There is also another local bundle (bundle-local-federated)

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
saltiyazan: It is true that `_configure_orc8r` already check if the container can be connected to, but if not `_configure_orc8r`  will call `defer()` and then return and the execution of `_on_magma_orc8r_orchestrator_pebble_ready` will continue and the `_create_orchestrator_admin_user()` function will be executed and will throw an exception. Of course `_on_magma_orc8r_orchestrator_pebble_ready` will be executed again because of the `defer()` but we still will have exceptions in the log.
I would either keep the redundant check, or to move the check it self from the  `_configure_orc8r` to  `_on_magma_orc8r_orchestrator_pebble_ready`

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
saltiyazan: I agree, addressed in the previous commit :)

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
saltiyazan: I agree, addressed in the previous commit

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
saltiyazan: sure I can use it instead :) could you please tell me what is the advantage of using it?

orchestrator-bundle/bundle-local.yaml
saltiyazan: good catch, thanks

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
gruyaume: In addition to emitting the event, it sets the "can connect" to true

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -452,10 +465,20 @@ def _on_magma_orc8r_orchestrator_pebble_ready(
...
         """
+        if not self._container.can_connect():
+            logger.info("Can't connect to container - Deferring")
+            event.defer()
+            return
         if not self._metrics_relation_created:
             self.unit.status = BlockedStatus("Waiting for metrics-endpoint relation to be created")
             event.defer()
             return
+        if not self._accessd_operator_relation_created_and_active:
```
gruyaume: Can we have two separate validations for the active and created? If the relation is not created, the status should be blocked but if the relation is not yet active the status should be waiting.

saltiyazan: done :)

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -452,10 +474,20 @@ def _on_magma_orc8r_orchestrator_pebble_ready(
...
+        if not self._container.can_connect():
+            logger.info("Can't connect to container - Deferring")
+            event.defer()
+            return
         if not self._metrics_relation_created:
             self.unit.status = BlockedStatus("Waiting for metrics-endpoint relation to be created")
             event.defer()
             return
+        if not self._accessd_operator_relation_created:
+            self.unit.status = BlockedStatus(
```
ghislainbourgeois, line 486: Shouldn't this be `WaitingStatus`?

gruyaume: Relations created use "Blocked" because they depend on user action (to relate X to Y in this case). And we use "Waiting" when we're user has nothing to do and things should resolve by themselves.

ghislainbourgeois: Ok, makes sense, thank you!

ghislainbourgeois (approved):

gruyaume (approved):

PR #73, opened 2022-08-01
Enable scaling for orc8r-certifier

ghislainbourgeois:
Move certificates from `StoredState` to the peer relation application data bag. This lets all the units use the same certificates.

Also updates the `tls-certificates-interface` library to support non-leader units receiving certificates.

### DIFF
--- a/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
+++ b/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
@@ -500,6 +500,8 @@ class MagmaOrc8rCertifierCharm(CharmBase):
             - admin_operator.pem
             - admin_operator.key.pem
             - admin_operator.pfx
+            - admin_operator_password
+            - bootstrapper.key
 
         Returns:
             None
--- a/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
+++ b/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
@@ -43,6 +43,7 @@ from ops.charm import (
     CharmBase,
     InstallEvent,
     PebbleReadyEvent,
+    RelationCreatedEvent,
     RelationJoinedEvent,
 )
 from ops.main import main
@@ -132,6 +133,9 @@ class MagmaOrc8rCertifierCharm(CharmBase):
             self.cert_bootstrapper.on.private_key_request,
             self._on_bootstrapper_private_key_request,
         )
+        self.framework.observe(
+            self.on.replicas_relation_created, self._on_replicas_relation_created
+        )
 
     @property
     def _root_certs_are_stored(self) -> bool:
@@ -194,15 +198,6 @@ class MagmaOrc8rCertifierCharm(CharmBase):
             return True
         except psycopg2.OperationalError:
             return False
-
-    @property
-    def _replicas_relation_created(self) -> bool:
-        """Checks whether replicas relation is created.
-
-        Returns:
-            bool: Whether required relation
-        """
-        return self._relation_created("replicas")
 
     @property
     def _pebble_layer(self) -> Layer:
@@ -321,8 +316,11 @@ class MagmaOrc8rCertifierCharm(CharmBase):
             self.unit.status = BlockedStatus("Config 'domain' is not valid")
             event.defer()
             return
-        if not self._replicas_relation_created:
-            self.unit.status = WaitingStatus("Waiting for peer relation to be created")
+        self._write_metricsd_config_file()
+
+    def _on_replicas_relation_created(self, event: RelationCreatedEvent) -> None:
+        if not self._container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for container to be ready")
             event.defer()
             return
         if not self._certificates_are_generated():
@@ -331,7 +329,6 @@ class MagmaOrc8rCertifierCharm(CharmBase):
                 event.defer()
                 return
             self._generate_application_certificates()
-        self._write_metricsd_config_file()
         self._push_application_certificates()
 
     def _on_magma_orc8r_certifier_pebble_ready(
--- a/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
+++ b/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
@@ -585,7 +585,10 @@ class MagmaOrc8rCertifierCharm(CharmBase):
         Returns:
             bool: Whether certificates have been generated.
         """
-        app_data = self.model.get_relation("replicas").data[self.app]  # type: ignore[union-attr]
+        replicas = self.model.get_relation("replicas")
+        if not replicas:
+            return False
+        app_data = replicas.data[self.app]
         return (
             app_data.get("certifier_pem")
             and app_data.get("certifier_key")  # noqa: W503
--- a/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
+++ b/orchestrator-bundle/orc8r-certifier-operator/src/charm.py
@@ -45,7 +45,6 @@ from ops.charm import (
     PebbleReadyEvent,
     RelationJoinedEvent,
 )
-from ops.framework import StoredState
 from ops.main import main
 from ops.model import (
     ActiveStatus,
@@ -81,8 +80,6 @@ class MagmaOrc8rCertifierCharm(CharmBase):
     PROMETHEUS_URL = "http://orc8r-prometheus:9090"
     ALERTMANAGER_URL = "http://orc8r-alertmanager:9093"
 
-    _stored = StoredState()
-
     def __init__(self, *args):
         """Initializes all events that need to be observed."""
         super().__init__(*args)
@@ -95,7 +92,6 @@ class MagmaOrc8rCertifierCharm(CharmBase):
         self.provided_relation_name = list(self.meta.provides.keys())[0]
         self._container = self.unit.get_container(self._container_name)
         self._db = pgsql.PostgreSQLClient(self, "db")
-        self._stored.set_default(admin_operator_password="")
         self._service_patcher = KubernetesServicePatch(
             charm=self,
             ports=[ServicePort(name="grpc", port=9180, targetPort=9086)],
@@ -198,6 +194,15 @@ class MagmaOrc8rCertifierCharm(CharmBase):
             return True
         except psycopg2.OperationalError:
             return False
+
+    @property
+    def _replicas_relation_created(self) -> bool:
+        """Checks whether replicas relation is created.
+
+        Returns:
+            bool: Whether required relation
+        """
+        return self._relation_created("replicas")
 
     @property
     def _pebble_layer(self) -> Layer:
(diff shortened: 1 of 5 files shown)


orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -320,8 +325,18 @@ def _on_install(self, event: InstallEvent) -> None:
...
+            return
+        if not self._certificates_are_generated():
+            if not self.unit.is_leader():
+                self.unit.status = WaitingStatus("Waiting for leader to generate certificates")
+                event.defer()
+                return
+            self._generate_application_certificates()
         self._write_metricsd_config_file()
-        self._generate_application_certificates()
+        self._push_application_certificates()
```
gruyaume, line 332: Doing this action for both the leader and non-leaders is redundant right? Here we are using k8s storage that is mounted on both units so we are pushing the same files multiple times.

ghislainbourgeois: I was unsure and validated, but the storage is per-unit. So we do need to push the certificates.

gruyaume: Ok good to know!

gruyaume (approved):

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -316,8 +321,18 @@ def _on_install(self, event: InstallEvent) -> None:
             self.unit.status = BlockedStatus("Config 'domain' is not valid")
             event.defer()
             return
+        if not self._replicas_relation_created:
```
saltiyazan: what happens when we are using one unit only? how is this condition passed?
(More of a question than a review comment) :)

ghislainbourgeois: From what I understand, the peer relation is created and joined automatically by Juju. In my testing, I have seen problems where this code runs before it is created, but once the deferred event is retried, the relation has been created. The integration test validates this behavior by first deploying only one instance, waiting for it to be `Active-Idle` and then scaling up the application.

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -167,12 +172,12 @@ def test_given_pebble_ready_when_install_then_admin_application_certificates_are
...
-        private_key_3 = "whatever private key 3"
-        certificate = "whatever certificate"
-        ca_certificate = "whatever ca certificate"
-        pfx_package = "whatever pfx package content"
+        private_key_1 = b"whatever private key 1"
+        private_key_2 = b"whatever private key 2"
+        private_key_3 = b"whatever private key 3"
+        certificate = b"whatever certificate"
+        ca_certificate = b"whatever ca certificate"
+        pfx_package = b"whatever pfx package content"
```
sanchezfdezjavier, line 180: Why are we using byte-strings? I guess that is for encoding/decoding, is that right?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
sanchezfdezjavier: So Juju considers the only unit to be a peer itself? Is that correct?

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
ghislainbourgeois: Yes, the functions we are patching are returning byte-strings. It accidentally worked before because we were not doing encoding/decoding.

sanchezfdezjavier: Cool, thanks!

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
ghislainbourgeois: Yes, a standalone unit will still get this relation, and it will contain the app data bag and the unit data bag. Once another unit is created, it will join this relation.

saltiyazan: > So Juju considers the only unit to be a peer itself? Is that correct?

That's what I'm trying to understand as well, unfortunately I haven't found an explanation on this in the juju docs yet.

saltiyazan: > From what I understand, the peer relation is created and joined automatically by Juju. In my testing, I have seen problems where this code runs before it is created, but once the deferred event is retried, the relation has been created. The integration test validates this behavior by first deploying only one instance, waiting for it to be `Active-Idle` and then scaling up the application.

Cool thanks :)

saltiyazan: > Yes, a standalone unit will still get this relation, and it will contain the app data bag and the unit data bag. Once another unit is created, it will join this relation.

Cool, thanks :)

sanchezfdezjavier: Got it, thanks!

ghislainbourgeois: Yes, the documentation is not super clear on this, it could be improved. It is however documented in the code: https://github.com/juju/juju/blob/fd886a18e169e4741f557d6b574bba06ff0b190d/doc/entity-creation.md

gruyaume: One way to not have to validate that the replicas relation is created is by having the `_write_metricsd_config_file` and the  `_push_application_certificates` methods in a handler for the replicas relation created event. Here we do all the init actions in the on install event.

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -60,3 +60,15 @@ async def test_build_and_deploy(self, ops_test, setup, build_and_deploy):
...
+
+    async def test_build_and_deploy_and_scale(self, ops_test, setup, build_and_deploy):
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
+        await ops_test.model.applications[APPLICATION_NAME].scale(2)
+
+        await ops_test.model.wait_for_idle(
+            apps=[APPLICATION_NAME], status="active", timeout=1000, wait_for_units=2
+        )
+
+        await ops_test.model.applications[APPLICATION_NAME].scale(1)
```
gruyaume, line 73: I think we should separate the upscale and downscale tests in two different tests

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -210,6 +215,86 @@ def test_given_pebble_ready_when_install_then_admin_application_certificates_are
             source=private_key_3,
         )
 
+    @patch("ops.model.Container.exists")
+    @patch("ops.model.Container.push")
+    def test_given_pebble_ready_when_install_and_unit_is_not_leader_then_status_is_waiting(
```
gruyaume:
```suggestion
    def test_given_pebble_ready_and_unit_is_not_leader_when_install_then_status_is_waiting(
```

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -210,6 +215,86 @@ def test_given_pebble_ready_when_install_then_admin_application_certificates_are
...
+        self.harness.charm.on.install.emit()
+
+        assert self.harness.charm.unit.status == WaitingStatus(
+            "Waiting for leader to generate certificates"
+        )
+        patch_push.assert_not_called()
+
+    @patch("ops.model.Container.exists")
+    @patch("ops.model.Container.push")
+    def test_given_pebble_ready_when_install_and_unit_is_not_leader_and_certs_are_generated_then_admin_application_certificates_are_pushed(  # noqa: E501
```
gruyaume:
```suggestion
    def test_given_pebble_ready_and_unit_is_not_leader_and_certs_are_generated_when_install_then_admin_application_certificates_are_pushed(  # noqa: E501
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -477,7 +492,7 @@ def _update_relations(self) -> None:
             )
 
     def _generate_application_certificates(self) -> None:
-        """Generates application certificates and pushes them to the workload container.
+        """Generates application certificates.
 
         The following certificates/keys are created:
             - certifier.pem
```
gruyaume, line 495: `bootstrapper.key` is also created

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
ghislainbourgeois: Fixed

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
ghislainbourgeois: Fixed

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
ghislainbourgeois: The thing is that the scale down test will completely contain the scale up test, as I will need to first scale up the number of instances before scaling down. It will also make the tests longer to run. 

Another way of doing the tests would be to move the scaling up in a fixture used by both scaling up and scaling down tests, but the scaling up tests would now be empty.

Do you have an idea on how to separate them cleanly in that context?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
ghislainbourgeois: Fixed

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
gruyaume: It's possible to deploy an app with a certain number of units at deploy time (using `-n`). So we'd use that to deploy the app with (Ex. 2 units) and scale it down to 1.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
ghislainbourgeois: I moved the `_push_application_certificates` and certificates generation to the relation created event. The `_write_metricsd_config_file` does not need the relation, so I left it there.

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
ghislainbourgeois: I played a bit with the tests, and it turns out that anyway, they are run sequentially, So splitting them is not longer. However, I realized that the scaling down test was not properly validating that it was indeed back to 1 unit, and when I fixed that, I realized that the scaling down does not work on Kubernetes due to a known bug. I put the test on `xfail` for now with the reason, so this will break when the bug is fixed, so we can remove the `xfail` then.

gruyaume: Ahh we should probably separate them so that they don't depend on each other, but that's a problem for another day.

PR #83, opened 2022-08-12
Uses the tls-certificates interface `v1`

gruyaume:
# Description

- Uses the tls-certificates interface `v1`
- Decouples private key creation from certificate creation
- This PR is dependent on this other [PR](https://github.com/canonical/tls-certificates-operator/pull/4) to be merged.

# Checklist:


orchestrator-bundle/orc8r-certifier-operator/pyproject.toml
```diff
@@ -26,7 +26,7 @@ copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
 [tool.mypy]
 pretty = true
 python_version = 3.8
-mypy_path = "$MYPY_CONFIG_FILE_DIR/src:$MYPY_CONFIG_FILE_DIR/lib:$MYPY_CONFIG_FILE_DIR/tests/unit"
+mypy_path = "$MYPY_CONFIG_FILE_DIR/src:$MYPY_CONFIG_FILE_DIR/lib/charms/magma_orc8r_certifier:$MYPY_CONFIG_FILE_DIR/tests/unit"
```
ghislainbourgeois: This change breaks mypy checking of imports.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -12,35 +12,51 @@
 
 import ops.lib
 import psycopg2  # type: ignore[import]
-from charms.magma_orc8r_certifier.v0.cert_admin_operator import (
+from charms.magma_orc8r_certifier.v0.cert_admin_operator import (  # type: ignore[import]
```
ghislainbourgeois: All the `type: ignore[import]` comments added in this PR should be removed after fixing pyproject.toml.

ghislainbourgeois (changes requested):
I did a quick first pass but did not go through all the code yet.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-certifier-operator/pyproject.toml
gruyaume: Fixed

orchestrator-bundle/orc8r-bundle/README.md
```diff
@@ -26,11 +26,12 @@ applications:
...
   tls-certificates-operator:
     options:
-      certificate: <your base64 encoded certificate>
-      private-key: <your base64 encoded private key>
-      ca-certificate: <your base64 encoded ca certificate>
+      generate-self-signed-certificates: true
+      ca-common-name: rootca.<your domain name>
 ```
 
+> **Warning**: This configuration is unsecure because it uses self-signed certificates.
```
ghislainbourgeois, line 33: I think it would be better to provide both ways (self-signed vs passing certificates), this would let the user know the alternative. If not, the user will have to know to go look up the documentation for `tls-certificates-operator`.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -102,7 +109,19 @@ def __init__(self, *args):
...
         self.framework.observe(
-            self.certificates.on.certificate_available, self._on_certificate_available
+            self.tls_certificates_requirer.on.certificate_available, self._on_certificate_available
+        )
+        self.framework.observe(
+            self.tls_certificates_requirer.on.certificate_expiring, self._on_certificate_expiring
+        )
+        self.framework.observe(
+            self.tls_certificates_requirer.on.certificate_expiring, self._on_certificate_expiring
+        )
```
ghislainbourgeois: This line is doubled.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -117,55 +136,89 @@ def __init__(self, *args):
...
 
     @property
-    def _root_certs_are_stored(self) -> bool:
-        """Returns whether root certificate are stored.
+    def _root_certificates_are_pushed(self) -> bool:
+        """Returns whether root certificate are pushed to workload.
 
         Returns:
-            bool: Whether root certificate are stored.
+            bool: True/False
```
ghislainbourgeois: I prefer the previous line here, we already know that a `bool` is `True` or `False`.
```suggestion
            bool: Whether root certificate are pushed to workload.
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -117,55 +136,89 @@ def __init__(self, *args):
...
+        """
+        if not self._container.exists(f"{self.BASE_CERTIFICATES_PATH}/certifier.key"):
+            logger.info("Application private key is not pushed")
+            return False
+        if not self._container.exists(f"{self.BASE_CERTIFICATES_PATH}/admin_operator.key.pem"):
+            logger.info("Admin operator private key is not pushed")
+            return False
+        if not self._container.exists(f"{self.BASE_CERTIFICATES_PATH}/bootstrapper.key"):
+            logger.info("Bootstrapper private key is not pushed")
+        return True
```
ghislainbourgeois: Is it intentional that we return `True` in this case?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -312,27 +368,148 @@ def _on_install(self, event: InstallEvent) -> None:
...
             event.defer()
             return
-        if not self._domain_config_is_valid:
-            self.unit.status = BlockedStatus("Config 'domain' is not valid")
-            event.defer()
-            return
-        self._write_metricsd_config_file()
+        self._push_metricsd_config_file()
+
+    def _on_config_changed(self, event: ConfigChangedEvent):  # noqa: C901
```
ghislainbourgeois: I am not a fan of ignoring `C901`, I think it usually is right. In this case, we could split this method in two:

```
    def _on_config_changed(self, event: ConfigChangedEvent):
         if self.unit.is_leader():
             self._on_leader_config_change()
         else:
             self._on_non_leader_config_change()
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -378,14 +585,25 @@ def _on_certificates_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+            event.defer()
+            return
+        if not self._root_csr:
+            self.unit.status = WaitingStatus("Waiting for root CSR to be generated")
             event.defer()
             return
-        self.certificates.request_certificate(
-            cert_type="server",
-            common_name=f"*.{domain_name}",
+        self._request_certificate_based_on_stored_csr()
```
ghislainbourgeois, line 587: This will be called whenever the `tls-certificates` provider is scaled up, will this be a problem?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -312,27 +368,148 @@ def _on_install(self, event: InstallEvent) -> None:
...
+                else:
+                    self.unit.status = BlockedStatus(
+                        "Waiting for tls-certificates relation to be created"
+                    )
+            if not self._application_certificates_are_stored:
+                self._generate_application_certificates()
+                self._push_application_certificates()
+            if not self._stored_application_certificate_matches_config:
+                self._generate_application_certificates()
+                self._push_application_certificates()
```
ghislainbourgeois:
```suggestion
            if not self._application_certificates_are_stored or not self._stored_application_certificate_matches_config:
                self._generate_application_certificates()
                self._push_application_certificates()
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -359,16 +535,47 @@ def _on_magma_orc8r_certifier_pebble_ready(
...
+            raise RuntimeError("No stored private key")
+        csr = generate_csr(
+            private_key=self._root_private_key.encode(), subject=f"*.{self._domain_config}"
+        )
+
+        self._store_root_csr(csr.decode())
+        logger.info("Generated CSR for root certificate")
+
     def _on_certificates_relation_joined(self, event: RelationJoinedEvent) -> None:
         """Juju event triggered when pebble is ready.
```
ghislainbourgeois: Can you fix the docstring?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -399,23 +617,33 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
-                path=f"{self.BASE_CERTS_PATH}/rootCA.pem", source=certificate_data["ca"]
-            )
-            self._container.push(
-                path=f"{self.BASE_CERTS_PATH}/controller.crt", source=certificate_data["cert"]
-            )
-            self._container.push(
-                path=f"{self.BASE_CERTS_PATH}/controller.key", source=certificate_data["key"]
-            )
+        if event.certificate_signing_request != self._root_csr:
+            logger.info("Certificate's CSR doesn't match stored root CSR")
```
ghislainbourgeois, line 610: In this case should we change the status to `BlockedStatus`?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -491,112 +719,427 @@ def _update_relations(self) -> None:
...
+            source=self._admin_operator_certificate,
+        )
+        self._container.push(
+            path=f"{self.BASE_CERTIFICATES_PATH}/admin_operator.pfx",
+            source=base64.b64decode(self._admin_operator_pfx),
+        )
+        logger.info("Pushed application certificates")
+
+    def _push_application_private_keys(self) -> None:
+        """Pushes application certificates to the workload container.
```
ghislainbourgeois:
```suggestion
        """Pushes application private keys to the workload container.
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -491,112 +719,427 @@ def _update_relations(self) -> None:
...
         )
         self._container.push(
-            path=f"{self.BASE_CERTS_PATH}/admin_operator.pfx",
-            source=base64.b64decode(app_data.get("admin_operator_pfx").encode()),
+            path=f"{self.BASE_CERTIFICATES_PATH}/controller.crt", source=self._root_certificate
         )
+        logger.info("Pushed root certificates")
+
+    def _push_root_private_key(self) -> None:
+        """Pushes root certificates to workload container.
```
ghislainbourgeois:
```suggestion
        """Pushes root private key to workload container.
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -491,112 +719,427 @@ def _update_relations(self) -> None:
...
+        """
+        if not self._application_certificate:
+            raise RuntimeError("Application certificates not stored")
+        application_certificate = x509.load_pem_x509_certificate(
+            data=self._application_certificate.encode()
         )
+        for subject in application_certificate.subject:
+            if subject.value == f"certifier.{self._domain_config}":
+                return True
+        logger.info("Stored application certificates dont match config")
```
ghislainbourgeois:
```suggestion
        logger.info("Stored application certificates does not match config")
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -491,112 +719,427 @@ def _update_relations(self) -> None:
...
+        """
+        if not self._root_csr:
+            raise RuntimeError("No stored root CSR")
+        csr_object = x509.load_pem_x509_csr(data=self._root_csr.encode())
+        try:
+            assert f"*.{self._domain_config}" == list(csr_object.subject)[0].value
+            return True
+        except AssertionError:
+            logger.info("Root CSR subject doesn't match with config")
+            return False
```
ghislainbourgeois: Is there a reason to use an assertion here instead of `if`?

ghislainbourgeois (changes requested):

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -29,7 +29,10 @@ async def _deploy_tls_certificates_operator(ops_test):
         await ops_test.model.deploy(
             "tls-certificates-operator",
             application_name="tls-certificates-operator",
-            config={"generate-self-signed-certificates": True},
+            config={
+                "generate-self-signed-certificates": True,
+                "ca-common-name": DOMAIN
```
Gmerold: According to the `README` above, shouldn't that be `f"rootca.{DOMAIN}"`?

orchestrator-bundle/orc8r-certifier-operator/tests/unit/certificates.py
```diff
@@ -1,185 +1,208 @@
...
 
 
 def generate_certificate(
     csr: bytes,
     ca: bytes,
     ca_key: bytes,
-    validity: int = 365,
-    alt_names: list = None,
+    ca_key_password: Optional[bytes] = None,
+    validity: int = 24 * 365,
```
Gmerold:
```suggestion
    validity: int = 365 * 24,
```
I know it's a cosmetic change, but at first glance it looks like the default cert validity was 24 years ;)

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: I'd say yes, but anyway, `return False` is missing from the last `if`. Isn't it?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -270,8 +327,7 @@ def _domain_config_is_valid(self) -> bool:
         Returns:
             bool: Whether the domain is a valid one.
         """
-        domain = self.model.config.get("domain")
-        if not domain:
+        if not self._domain_config:
```
Gmerold, line 317: You might wanna use `validators.domain` for stronger validation.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: +1

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -359,16 +535,47 @@ def _on_magma_orc8r_certifier_pebble_ready(
...
 
+    def _generate_root_csr(self) -> None:
+        """Generates a CSR with the domain name in the Juju config.
+
+        Returns:
+            None
+        """
+        domain_name = self._domain_config
+        peer_relation = self.model.get_relation("replicas")
+        if not domain_name:
```
Gmerold: Is this check needed?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -359,16 +535,47 @@ def _on_magma_orc8r_certifier_pebble_ready(
...
+            raise RuntimeError("No domain name set")
+        if not peer_relation:
+            raise RuntimeError("No peer relation")
+        if not self._domain_config_is_valid:
```
Gmerold, line 558: Is this check needed? 
BTW, considering the current implementation of this check, it does the same as the `if not domain_name` above.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -359,16 +535,47 @@ def _on_magma_orc8r_certifier_pebble_ready(
...
+            raise ValueError("Domain config is not valid")
+        if not self._root_private_key:
```
Gmerold: Is this check needed?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -399,23 +617,33 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
         Returns:
             None
         """
-        logger.info("Certificate is available")
         if not self._container.can_connect():
             logger.info("Cant connect to container - Won't push certificates to workload")
```
Gmerold: Don't we wanna change the status to `WaitingStatus`?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -84,11 +86,16 @@ class MagmaOrc8rCertifierCharm(CharmBase):
     def __init__(self, *args):
         """Initializes all events that need to be observed."""
         super().__init__(*args)
-        self.certificates = TLSCertificatesRequires(self, "certificates")
-        self.cert_admin_operator = CertAdminOperatorProvides(self, "cert-admin-operator")
-        self.cert_certifier = CertCertifierProvides(self, "cert-certifier")
-        self.cert_controller = CertControllerProvides(self, "cert-controller")
-        self.cert_bootstrapper = CertBootstrapperProvides(self, "cert-bootstrapper")
+        self.tls_certificates_requirer = TLSCertificatesRequiresV1(self, "certificates")
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
```
sanchezfdezjavier: Is there any reason why this `observe` is here alone, instead of at the bottom with the rest of them?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: I don't think we need to. `BlockedStatus` typically indicates need for user's intervention. 
If I'm not mistaken, this check can fail only if someone changes the `domain` or deletes the `CSR` in the time between requesting and issuing the cert.
I'd increase the log level though.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -436,7 +664,7 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
         if self.unit.is_leader():
             event.database = self.DB_NAME  # type: ignore[attr-defined]
 
-    def _write_metricsd_config_file(self) -> None:
+    def _push_metricsd_config_file(self) -> None:
```
Gmerold, line 655: Shouldn't this file also contain keys for Prometheus Configurer and Alertmanager Configurer?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -312,27 +368,148 @@ def _on_install(self, event: InstallEvent) -> None:
...
-                self.unit.status = WaitingStatus("Waiting for leader to generate certificates")
-                event.defer()
-                return
-            self._generate_application_certificates()
+        self._generate_root_private_key()
+        self._generate_application_private_keys()
+        self._push_application_private_keys()
+        self._push_root_private_key()
+
+    def _on_replicas_relation_joined(self, event: RelationJoinedEvent):
```
sanchezfdezjavier: Is there any reason behind not having a docstring for this method?

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: We will have to think a bit more about how people will provide signed certificates of their own. There are multiple options:
1. The user provides certifier with a private key in the config and the tls-certficates-operator with both the cert and ca cert.
2. The user retrieves the CSR from certifier and then generates/provides certs that they insert into tls-cerrificates-operator

I think option 1 is the way to go but I didn't want to bundle this discussion with this already super large PR.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
saltiyazan: I agree too, but something more like 
```
    def _on_config_changed(self, event: ConfigChangedEvent):
        if not self._container.can_connect():
               self.unit.status = WaitingStatus("Waiting for container to be ready")
               event.defer()
               return
         if self.unit.is_leader():
             self._on_leader_config_change()
         else:
             self._on_non_leader_config_change()
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -312,27 +368,148 @@ def _on_install(self, event: InstallEvent) -> None:
...
+                event.defer()
+                return
+            if not self._root_csr_is_stored:
+                self.unit.status = WaitingStatus("Waiting for leader to generate root csr")
+                event.defer()
+                return
+            if not self._stored_root_csr_matches_config:
+                self.unit.status = WaitingStatus("Waiting for leader to generate new root csr")
+                event.defer()
+                return
```
saltiyazan: like above this also can be one condition, the root csr is new in both cases, isn't it? or do we need to distinguish between the two?
```suggestion
            if not self._root_csr_is_stored or not self._stored_root_csr_matches_config:
                self.unit.status = WaitingStatus("Waiting for leader to generate a new root csr")
                event.defer()
                return
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: as far as I know, those are the only things it needs

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed now.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
ghislainbourgeois: But if this check fails, won't the charm be effectively blocked until a user intervenes to fix the issue?

Gmerold: Changing the `domain` will trigger `config_changed` event, in which case the charm should recover automatically. Not sure about deleting the CSR manually, but that seems like a wild thing to do, doesn't it?

ghislainbourgeois: Yes, good point

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: I'm not certain we want to add this project as a dependency, but I'm happy to steal their regex

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: `_stored_application_certificate_matches_config` raises if there is no stored certificate, hence why I placed them on 2 separate lines

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: `_stored_application_certificate_matches_config` raises if there is no stored certificate, hence why I placed them on 2 separate lines

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Yeah I think so, if in the future we call generate_root_csr without a domain name, we'd want it to error our properly.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Yes, we don't want to push an empty root private key. In the charm today this should never be triggered by nothing prevents us in the future from using this method and ending up pushing an empty private key without erroring out

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
ghislainbourgeois: In my suggested change, since the check for stored certificate is before, the second check will not be run in that case, because the first element of the `or` is `True`, so the rest does not matter.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Just used the domain_config_is_valid and removed the other.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Done

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Naa, moved it to a if

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: No, forgot about it but this method doesn't exist anymore

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
gruyaume: Sure fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Good point, fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Done

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: Fixed, now being called on relation created only.

orchestrator-bundle/orc8r-bundle/README.md
ghislainbourgeois: As discussed in the daily, I agree that it needs to be done separately from this PR.

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
```diff
@@ -107,42 +100,98 @@ def _service_is_running(self) -> bool:
...
+        if self.unit.is_leader():
+            bootstrapper_key = generate_private_key()
+            self._store_bootstrapper_private_key(bootstrapper_key.decode())
+        else:
+            if not self._bootstrapper_private_key_is_stored:
+                self.unit.status = WaitingStatus(
+                    "Waiting for leader to generate bootstrapper private key"
+                )
+                event.defer()
+                return
```
sanchezfdezjavier: A little bit less nesting?
```suggestion
        elif not self._bootstrapper_private_key_is_stored:
+            self.unit.status = WaitingStatus(
+                "Waiting for leader to generate bootstrapper private key"
+            )
+            event.defer()
+            return
```

gruyaume: Done

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
saltiyazan: > `_stored_application_certificate_matches_config` raises if there is no stored certificate, hence why I placed them on 2 separate lines

Was this reply intended to this comment? if so I don't fully get it :D

gruyaume: Yeah it's fixed now anywaybut yes what I meant was that the second `if` would raise an exception if the root csr isn't stored but like Ghislain mentioned above, we will never get in that situation given the first if would catch it.

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
```diff
@@ -107,42 +100,97 @@ def _service_is_running(self) -> bool:
...
+        replicas = self.model.get_relation("replicas")
+        if not replicas:
+            return None
+        return replicas.data[self.app].get("bootstrapper_private_key", None)
 
     @property
-    def _cert_bootstrapper_relation_created(self) -> bool:
-        """Returns whether cert-bootstrapper relation is created.
+    def _bootstrapper_private_key_is_stored(self) -> bool:
+        """Returns whether certificates are stored in relation data.
```
saltiyazan: it seems to me as it checks if the `private key` is stored not `certificates`

```suggestion
        """Returns whether bootstrapper private key is stored in peer relation data.
```

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
```diff
@@ -107,42 +100,97 @@ def _service_is_running(self) -> bool:
...
 
         Returns:
-            bool: True/False
+            bool: Whether certificates have been generated.
```
saltiyazan: same as above
```suggestion
            bool: Whether private key has been stored.
```

orchestrator-bundle/orc8r-bootstrapper-operator/metadata.yaml
```diff
@@ -32,7 +32,6 @@ provides:
   magma-orc8r-bootstrapper:
     interface: magma-orc8r-bootstrapper
 
-requires:
-  cert-bootstrapper:
-    interface: cert-bootstrapper
-    limit: 1
+peers:
+  replicas:
+    interface: orc8r-certifier-replica
```
saltiyazan: shouldn't it be `orc8r-bootstrapper-replica` ?

gruyaume: Fixed, good catch

orchestrator-bundle/orc8r-bootstrapper-operator/metadata.yaml
```diff
@@ -32,7 +32,6 @@ provides:
   magma-orc8r-bootstrapper:
     interface: magma-orc8r-bootstrapper
 
-requires:
-  cert-bootstrapper:
-    interface: cert-bootstrapper
```
saltiyazan, line 37: Could you please remove this relation from `bundle.yaml`, `bundle-local.yaml` and `bundle-local-federated.yaml`?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -248,6 +280,15 @@ def _certificates_relation_created(self) -> bool:
         """
         return self._relation_created("certificates")
 
+    @property
+    def _replicas_relation_created(self) -> bool:
+        """Returns whether the replicas  relation is created.
```
sanchezfdezjavier:
```suggestion
        """Returns whether the replicas relation is created.
```

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -29,7 +29,10 @@ async def _deploy_tls_certificates_operator(ops_test):
         await ops_test.model.deploy(
             "tls-certificates-operator",
             application_name="tls-certificates-operator",
-            config={"generate-self-signed-certificates": True},
+            config={
+                "generate-self-signed-certificates": True,
+                "ca-common-name": f"rootca.{DOMAIN}",
```
saltiyazan, line 34: I think this should be added to all integration tests that deploy `tls-certificate-operator`

- `nms-magmalte`
- `nms-nginx-proxy`
- `orc8r-nginx`
- `orc8r-metricsd`
- `orc8r-orchestrator`

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -312,23 +363,142 @@ def _on_install(self, event: InstallEvent) -> None:
...
+        if not self._domain_config_is_valid:
+            self.unit.status = BlockedStatus("Config 'domain' is not valid")
+            event.defer()
+            return
+        if not self._root_private_key_is_stored:
+            self.unit.status = WaitingStatus("Waiting for root private key to be generated")
+            event.defer()
+            return
+        if not self._application_private_keys_are_stored:
+            self.unit.status = WaitingStatus("Waiting for application private key to be generated")
```
sanchezfdezjavier:
```suggestion
            self.unit.status = WaitingStatus("Waiting for application private keys to be generated")
```

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
```diff
@@ -107,42 +88,81 @@ def _service_is_running(self) -> bool:
...
+        """Stores bootstrapper private key in peer relation data."""
+        peer_relation = self.model.get_relation("replicas")
+        if not peer_relation:
+            raise RuntimeError("No peer relation")
+        peer_relation.data[self.app].update({"bootstrapper_private_key": private_key})
 
-    def _on_magma_orc8r_bootstrapper_pebble_ready(
-        self, event: Union[PebbleReadyEvent, PrivateKeyAvailableEvent]
-    ) -> None:
+    def _on_magma_orc8r_bootstrapper_pebble_ready(self, event: Union[PebbleReadyEvent]) -> None:
```
ghislainbourgeois:
```suggestion
    def _on_magma_orc8r_bootstrapper_pebble_ready(self, event: PebbleReadyEvent) -> None:
```

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
```diff
@@ -151,19 +171,17 @@ def _on_magma_orc8r_bootstrapper_pebble_ready(
...
-        if not self._certs_are_stored:
-            self.unit.status = WaitingStatus("Waiting for certs to be available")
+        if not self._bootstrapper_private_key_is_pushed:
+            self.unit.status = WaitingStatus("Waiting for bootstrapper private key to be pushed")
             event.defer()
             return
         self._configure_pebble(event)
 
-    def _configure_pebble(self, event: Union[PebbleReadyEvent, PrivateKeyAvailableEvent]) -> None:
+    def _configure_pebble(self, event: Union[PebbleReadyEvent]) -> None:
```
ghislainbourgeois:
```suggestion
    def _configure_pebble(self, event: PebbleReadyEvent) -> None:
```

ghislainbourgeois (changes requested):

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
```diff
@@ -107,42 +88,81 @@ def _service_is_running(self) -> bool:
...
+    def _on_install(self, event: InstallEvent) -> None:
+        """Triggered on charm install.
 
         Args:
-            relation_name (str): Relation name
+            event: Juju event
 
         Returns:
-            bool: True/False
+            None
```
sanchezfdezjavier, line 128: Which criteria are we following for including `Args` and `Returns` sections in docstrings? I see we are removing those from some methods.

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
gruyaume: Fixed

orchestrator-bundle/orc8r-bootstrapper-operator/src/charm.py
gruyaume: I removed those that were trivial (ex. returning a bool) and kept those where there was value in explaining what the return values were.

orchestrator-bundle/orc8r-bootstrapper-operator/metadata.yaml
gruyaume: Done

ghislainbourgeois (approved):

orchestrator-bundle/orc8r-bootstrapper-operator/tests/unit/test_charm.py
```diff
@@ -127,15 +170,75 @@ def test_given_magma_orc8r_bootstrapper_service_not_running_when_magma_orc8r_boo
...
-            path="/var/opt/magma/certs/bootstrapper.key", source="whatever"
+        relation_data = self.harness.get_relation_data(
+            relation_id=peer_relation_id, app_or_unit=self.harness.charm.app.name
+        )
+
+        self.harness.charm._on_install(event=event)
+
+        serialization.load_pem_private_key(
+            relation_data["bootstrapper_private_key"].encode(), password=None
+        )
```
sanchezfdezjavier, line 190: An empty private key would still pass the test? I guess `load_pem_private_key` is the method from `cryptography` module you were talking about so that's also checked. Is that right?

orchestrator-bundle/orc8r-bootstrapper-operator/tests/unit/test_charm.py
```diff
@@ -127,15 +170,75 @@ def test_given_magma_orc8r_bootstrapper_service_not_running_when_magma_orc8r_boo
...
+        event = Mock()
+
+        self.create_peer_relation_with_private_key(bootstrapper_private_key=False)
+
+        self.harness.charm._on_install(event=event)
+
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            WaitingStatus("Waiting for leader to generate bootstrapper private key"),
         )
```
sanchezfdezjavier: Aren't we either asserting the wrong thing or the test name is wrong?

sanchezfdezjavier: I'm not sure, but from the test above I guess the name should be something like `test_..._then_status_is_waiting`

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -118,54 +127,77 @@ def __init__(self, *args):
...
-            and self._container.exists(f"{self.BASE_CERTS_PATH}/controller.key")  # noqa: W503
-            and self._container.exists(f"{self.BASE_CERTS_PATH}/rootCA.pem")  # noqa: W503
-        )
+        if not self._container.exists(f"{self.BASE_CERTIFICATES_PATH}/controller.crt"):
+            logger.info("Root certificate is not pushed")
+            return False
+        if not self._container.exists(f"{self.BASE_CERTIFICATES_PATH}/rootCA.pem"):
+            logger.info("Root CA Certificate is not pushed")
+            return False
+        return True
```
sanchezfdezjavier, line 155: How about outsourcing this logic to a separate method, I see we're repeating it several times for each certificate

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -359,33 +530,69 @@ def _on_magma_orc8r_certifier_pebble_ready(
...
-        """Juju event triggered when pebble is ready.
+    def _generate_root_csr(self) -> None:
+        """Generates a CSR with the domain name in the Juju config.
+
+        Returns:
+            None
+        """
+        peer_relation = self.model.get_relation("replicas")
+        if not self._domain_config_is_valid:
+            raise ValueError("Domain config is not valid")
```
sanchezfdezjavier, line 559: Why are we rising instead of setting status as blocked? Wouldn't the charm need user intervention for setting up a new valid domain?

sanchezfdezjavier (changes requested):

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: I'm not sure I get what would be in this method?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
gruyaume: This is a helper method, we don't want it to set status. In theory, the raise will never occur because the method that actually handles the event will check if the relation exists.

orchestrator-bundle/orc8r-bootstrapper-operator/tests/unit/test_charm.py
gruyaume: No, `load_pem_private_key` would raise if the key is empty. Yes, the method is part of the cryptography library.

orchestrator-bundle/orc8r-bootstrapper-operator/tests/unit/test_charm.py
gruyaume: Good catch, fixed

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
sanchezfdezjavier: Got it, thanks

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
sanchezfdezjavier: Nevermind. It can't be done, my bad.

sanchezfdezjavier (approved):
LGTM

orchestrator-bundle/orc8r-bundle/README.md
```diff
@@ -26,11 +26,12 @@ applications:
       domain: <your domain name>
   tls-certificates-operator:
     options:
-      certificate: <your base64 encoded certificate>
-      private-key: <your base64 encoded private key>
-      ca-certificate: <your base64 encoded ca certificate>
+      generate-self-signed-certificates: true
+      ca-common-name: rootca.<your domain name>
```
saltiyazan, line 30: Does this affect the 'how-to' guide on magma's official documentation?
https://docs.magmacore.org/docs/next/orc8r/deploy_using_juju

orchestrator-bundle/orc8r-bundle/bundle.yaml
```diff
@@ -180,7 +180,6 @@ applications:
     charm: tls-certificates-operator
     channel: edge
     scale: 1
-    trust: true
```
saltiyazan, line 183: could you please remove 
```
- - orc8r-bootstrapper
  - orc8r-certifier
 ```

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: yes

orchestrator-bundle/orc8r-bundle/bundle.yaml
gruyaume: Done

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: OK, just asking, because in the `metricsd` charm these additional keys are needed.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: Yes, my point was that before the changes, private key availability was already assured by the time this piece of code could be executed, so I thought this wasn't needed.
In the current implementation there's only one place when it's checked.

orchestrator-bundle/orc8r-bundle/README.md
gruyaume: https://github.com/magma/magma/pull/13681

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: Yes, my point was that validity of the domain is already ensured at this point. We run `generate_root_csr` in `_on_certificate_expiring` and `_on_leader_config_changed`. Both mentioned functions validate domain config anyway.

gruyaume: I see the change is "outdated" here. You can look at the current implementation and tell me what you think. The general idea though was that this method doesn't know what was executed before. It's not great to have to do steps before calling a method if you don't want it it crash. Btw it'd still raise in this case but cleanlier and with a proper error message.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
Gmerold: Right, that was my question exactly in another convo: why do we need this check if it will never occur?

gruyaume: Same answer as in the other place. This method doesn't know what was called before. Right now it'd never raise **but** if you end up using it somewhere else in a couple of weeks, how would you know that you need to do X before invoking it.

saltiyazan (approved):

Gmerold (approved):

PR #90, opened 2022-09-05
Adds promote github actions to orchestrator-bundle charms

saltiyazan:
# Description

Adds promote github action to each of the charms in orchestrator-bundle, that will allow the promotion of charms to `beta` and later to other channels.

# Checklist:


Gmerold (changes requested):
Could we have just one workflow with additional parameter specifying charm (or bundle) name?
Also what about bundle promotion? I can't see the workflow for the bundle. Was that intended?

orchestrator-bundle/orc8r-bundle/bundle.yaml
```diff
@@ -7,17 +7,14 @@ description: |
...
   nms-nginx-proxy:
     charm: magma-nms-nginx-proxy
-    channel: edge
     scale: 1
     trust: true
   orc8r-accessd:
     charm: magma-orc8r-accessd
-    channel: edge
     scale: 1
     trust: true
```
Gmerold: Do we still need `trust: true` if we're not using `edge`?

saltiyazan:
> Could we have just one workflow with additional parameter specifying charm (or bundle) name? Also what about bundle promotion? I can't see the workflow for the bundle. Was that intended?

There is a [limitation](https://github.com/canonical/charming-actions/tree/main/release-charm#limitations) for the `releae-charm` action, it doesn't support releasing bundles :)

Now we have a single workflow with parameters :)

orchestrator-bundle/orc8r-bundle/bundle.yaml
saltiyazan: Fixed

Gmerold: That was actually a question, not a suggestion ;)
According to [this](https://juju.is/docs/olm/trust-an-application-with-a-credential), `trust` flag has nothing to do with channels; so I reckon we need all these after all :D

saltiyazan: Good point, fixed again :)

Gmerold (commented):
BTW, if I understood correctly, you'll be promoting charm immediately, correct? 
If not, we should probably wait with removing `channel: edge` until we have something in `beta`.

saltiyazan:
> BTW, if I understood correctly, you'll be promoting charm immediately, correct? If not, we should probably wait with removing `channel: edge` until we have something in `beta`.

Correct, I moved that to a separate PR so we have more control on when to merge it :)

Gmerold (approved):

ghislainbourgeois (approved):

PR #99, opened 2022-09-07
nms-magmalte implements grafana-auth lib as a provider of the authentication configuration

saltiyazan:
# Description

nms-magmalte implements grafan-auth lib and provides the authentication configuration needed (auth proxy).

# Checklist:


orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -513,6 +540,10 @@ def _generate_password() -> str:
         chars = string.ascii_letters + string.digits
         return "".join(secrets.choice(chars) for _ in range(12))
 
+    def _on_urls_available(self, event):
+        if event.urls:
+            self.grafana_url = event.urls[0]
```
gruyaume: Setting this attribute won't persist on subsequent events. You better use peer relation data. In this case, the relation is called "replicas"

gruyaume (changes requested):

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -361,3 +388,12 @@ def test_given_pebble_ready_when_certificate_available_then(self, patch_push):
...
             ]
         )
+
+    def test_given_grafana_auth_relation_when_urls_available_event_then_grafana_urls_are_stored_in_peer_data(  # noqa: E501
+        self,
+    ):
+        event = Mock()
+        event.urls = self.GRAFANA_URLS
+        self.harness.charm._on_grafana_urls_available(event=event)
+        url = self.harness.charm._grafana_url
```
gruyaume: Don't validate that a property is set (what if we change the property name in the future). Fetch the relation data and look into it for the value.

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -171,7 +191,7 @@ def test_given_relations_are_created_and_certs_are_stored_when_pebble_ready_then
                         "MAPBOX_ACCESS_TOKEN": "",
                         "MYSQL_DIALECT": "postgres",
                         "PUPPETEER_SKIP_DOWNLOAD": "true",
-                        "USER_GRAFANA_ADDRESS": "orc8r-user-grafana:3000",
+                        "USER_GRAFANA_ADDRESS": "auth-requirer:3000",
```
gruyaume: use the variable `GRAFANA_URLS`, here we don't want `USER_GRAFANA_ADDRESS` to specifically be "auth-requirer" but the content of that variable, whatever it ends up being.

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -513,6 +539,23 @@ def _generate_password() -> str:
         chars = string.ascii_letters + string.digits
         return "".join(secrets.choice(chars) for _ in range(12))
 
+    def _on_grafana_urls_available(self, event):
```
gruyaume: Can add typing for the event

orchestrator-bundle/orc8r-bundle/bundle.yaml.j2
```diff
@@ -168,6 +168,7 @@ applications:
     channel: edge
     options:
       web_external_url: "/grafana"
+      enable_auto_assign_org: false
```
gruyaume, line 180: I thought this was supposed to be true?

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -513,6 +542,28 @@ def _generate_password() -> str:
...
+    def _grafana_url(self) -> Optional[str]:
+        """Returns grafana url.
+
+        Returns:
+            str: grafana url
+        """
+        app_data = self.model.get_relation("replicas").data[self.app]  # type: ignore[union-attr]
+        if not app_data.get("grafana_url"):
+            return None
+        return app_data.get("grafana_url")
```
ghislainbourgeois:
```suggestion
        return app_data.get("grafana_url")
```

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -513,6 +542,28 @@ def _generate_password() -> str:
         chars = string.ascii_letters + string.digits
         return "".join(secrets.choice(chars) for _ in range(12))
 
+    def _on_grafana_urls_available(self, event: UrlsAvailableEvent):
+        """Triggered when grafana urls are available from relation data.
+
+        Args:
+            event: UrlsAvailableEvent.
+        """
+        if not self._grafana_url:
```
ghislainbourgeois: With this, it makes it impossible to change the grafana url after the fact. I don't think we expect it to really change, but I could be wrong. Do you have an opinion on this?

ghislainbourgeois (changes requested):

orchestrator-bundle/nms-magmalte-operator/src/charm.py
gruyaume: While this might "never" happen, it'd be pretty simple from our side to support a change of url, and it's a good practice for any relation data.

orchestrator-bundle/nms-magmalte-operator/src/charm.py
saltiyazan: Done

orchestrator-bundle/nms-magmalte-operator/src/charm.py
saltiyazan: Correct, I was worried that this event would be emitted quire frequently even if the urls don't change. 
But I fixed it now :)

orchestrator-bundle/orc8r-bundle/bundle.yaml.j2
saltiyazan: It is true by default.
Based on the [documentation](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#auto_assign_org) it should be set to false to create an organisation for each username, if it is true it will assign the user to the main org.
Looking at the [code](https://github.com/magma/magma/blob/35fbec2800f8b14ab2f71c3ce4406026a80f4ead/nms/server/grafana/handlers.ts#L192) of magma, I see that it expects grafana to create an org for each user, therefore it needs to be false.
Let me know if you think otherwise :)

gruyaume: Let's do whatever is done in the upstream Magma, I just thought it was set to `true` up there.

gruyaume (approved):

ghislainbourgeois (approved):

ghislainbourgeois (approved):

gruyaume (approved):

PR #106, opened 2022-09-22
Integrating Orc8r with Alertmanager Configurer

Gmerold:
# Description

- Integrates Magma Orc8r with Alertmanager Configurer.
- Updates DNS used by microk8s in self-hosted runners ([TELCO-270](https://warthogs.atlassian.net/browse/TELCO-270))

# Checklist:

- [X] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that validate the behaviour of the software
- [X] I validated that new and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
+++ b/orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
@@ -5,6 +5,7 @@
 """Collects runtime metrics from gateways and Orchestrator services."""
 
 import logging
+from typing import Optional
 
 from charms.observability_libs.v1.kubernetes_service_patch import (
     KubernetesServicePatch,
@@ -31,13 +32,11 @@ class MagmaOrc8rMetricsdCharm(CharmBase):
     BASE_CONFIG_PATH = "/var/opt/magma/configs/orc8r"
     REQUIRED_EXTERNAL_RELATIONS = [
         "alertmanager-k8s",
+        "alertmanager-configurer-k8s",
         "prometheus-k8s",
         "prometheus-configurer-k8s",
     ]
     REQUIRED_ORC8R_RELATIONS = ["magma-orc8r-orchestrator"]
-
-    # TODO: The various URL's should be provided through relationships.
-    ALERTMANAGER_CONFIGURER_URL = "http://orc8r-alertmanager:9101"
 
     def __init__(self, *args):
         """Uses the Orc8rBase library to manage events."""
@@ -84,10 +83,10 @@ class MagmaOrc8rMetricsdCharm(CharmBase):
         if not self._relations_created:
             event.defer()
             return
-        self._write_config_file()
         if not self._relations_ready:
             event.defer()
             return
+        self._write_config_file()
         if not self._container.can_connect():
             self.unit.status = WaitingStatus("Waiting for container to be ready")
             event.defer()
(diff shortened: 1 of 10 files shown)


sanchezfdezjavier (approved):
LGTM

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
```diff
@@ -31,14 +31,12 @@ class MagmaOrc8rMetricsdCharm(CharmBase):
...
     REQUIRED_EXTERNAL_RELATIONS = [
         "alertmanager-k8s",
+        "alertmanager-configurer-k8s",
         "prometheus-k8s",
         "prometheus-configurer-k8s",
     ]
     REQUIRED_ORC8R_RELATIONS = ["magma-orc8r-orchestrator"]
 
-    # TODO: The various URL's should be provided through relationships.
-    ALERTMANAGER_CONFIGURER_URL = "http://orc8r-alertmanager:9101"
```
gruyaume, line 40: :smile:

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
```diff
@@ -235,6 +241,27 @@ def _alertmanager_url(self) -> str:
...
         return f"http://{alertmanager_service_name}:9093"
 
+    @property
+    def _alertmanager_configurer_url(self) -> str:
+        """Returns the URL of the Alertmanager Configurer API.
+
+        Returns:
+            str: Alertmanager Configurer URL
+        """
+        alertmanager_configurer_relation = self.model.get_relation("alertmanager-configurer-k8s")
```
gruyaume, line 251: This makes me think that we should probably have a library to retrieve the info in this function. Out of scope of this PR though.

Gmerold: Technically, according to what I've recently learnt, we should extract pretty much all of our charms' business logic to libs. 
As for this particular case, if that's the only charm that requires this info (and I believe it is), perhaps creating an abstract method would make more sense (especially when there's also similar relation data published by `prometheus-k8s` and `alertmanager-k8s`)?
WDYT?

sanchezfdezjavier (approved):
LGTM

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
```diff
@@ -84,10 +82,10 @@ def _on_magma_orc8r_metricsd_pebble_ready(self, event: PebbleReadyEvent) -> None
         if not self._relations_created:
             event.defer()
             return
-        self._write_config_file()
         if not self._relations_ready:
             event.defer()
             return
+        self._write_config_file()
```
ghislainbourgeois: This is called before validating that we can connect to the container, won't that fail in some cases?

ghislainbourgeois (changes requested):

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
gruyaume: I don't think I agree about "extracting all of a charm business logic to libraries". However I believe that logic about relations (ex. tls) and external systems (ex. k8s services) should indeed be inside of libraries.

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
Gmerold: Probably. Fixed.

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
Gmerold: Yeah, I'm not a fan of extracting **everything** either. 
For COS integration logic, I've drafted a task in our backlog - [TELCO-269](https://warthogs.atlassian.net/browse/TELCO-269). Feel free to update it.

gruyaume (approved):

ghislainbourgeois (approved):

PR #109, opened 2022-09-27
Handles db_relation_departed event in orc8r_base_db lib

saltiyazan:
# Checklist:

### DIFF
--- a/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
+++ b/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
@@ -118,7 +118,7 @@ class Orc8rBase(Object):
         relation_joined_event = getattr(
             self.charm.on, f"{provided_relation_name_with_underscores}_relation_joined"
         )
-        self.framework.observe(pebble_ready_event, self._configure_orc8r)
+        self.framework.observe(pebble_ready_event, self._on_magma_orc8r_pebble_ready)
 
         if additional_environment_variables:
             self.additional_environment_variables = additional_environment_variables
@@ -129,7 +129,9 @@ class Orc8rBase(Object):
         self.framework.observe(
             self.db.on.database_relation_joined, self._on_database_relation_joined
         )
-        self.framework.observe(self.db.on.database_relation_broken, self._configure_orc8r)
+        self.framework.observe(
+            self.db.on.database_relation_broken, self._on_database_relation_broken
+        )
         self.framework.observe(relation_joined_event, self._on_relation_joined)
 
     @property
@@ -139,7 +141,7 @@ class Orc8rBase(Object):
             return False
         return True
 
-    def _configure_orc8r(self, event: Union[PebbleReadyEvent, RelationBrokenEvent]):
+    def _on_magma_orc8r_pebble_ready(self, event: Union[PebbleReadyEvent, RelationJoinedEvent]):
         if not self._db_relation_created:
             self.charm.unit.status = BlockedStatus("Waiting for db relation to be created")
             event.defer()
@@ -150,7 +152,7 @@ class Orc8rBase(Object):
             return
         self._configure_pebble(event)
 
-    def _configure_pebble(self, event: Union[PebbleReadyEvent, RelationBrokenEvent]):
+    def _configure_pebble(self, event: Union[PebbleReadyEvent, RelationJoinedEvent]):
         """Adds layer to pebble config if the proposed config is different from the current one."""
         if self.container.can_connect():
             self.charm.unit.status = MaintenanceStatus("Configuring pod")
@@ -197,6 +199,16 @@ class Orc8rBase(Object):
             event.database = self.DB_NAME  # type: ignore[attr-defined]
         else:
             event.defer()
+
+    def _on_database_relation_broken(self, event: RelationBrokenEvent):
+        """Event handler for database relation broken.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
+        Returns:
+            None
+        """
+        self.charm.unit.status = BlockedStatus("Waiting for db relation to be created")
 
     @property
     def _db_relation_ready(self) -> bool:
--- a/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
+++ b/orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
@@ -58,7 +58,12 @@ import logging
 
 import ops.lib
 import psycopg2  # type: ignore[import]
-from ops.charm import CharmBase, PebbleReadyEvent, RelationJoinedEvent
+from ops.charm import (
+    CharmBase,
+    PebbleReadyEvent,
+    RelationBrokenEvent,
+    RelationJoinedEvent,
+)
 from ops.framework import Object
 from ops.model import (
     ActiveStatus,
@@ -70,6 +75,7 @@ from ops.model import (
 )
 from ops.pebble import Layer
 from pgconnstr import ConnectionString  # type: ignore[import]
+from typing import Union
 
 # The unique Charmhub library identifier, never change it
 LIBID = "7e1096554dd649b78acd5f3187c017c8"
@@ -79,7 +85,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


Gmerold (changes requested):
Changes look good to me, but I think we could use some tests for the new behavior.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -191,6 +200,18 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent):
...
+    def _on_database_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Event handler for database relation broken.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
+
+        Returns:
+            None
+        """
+        self.charm.unit.status = BlockedStatus("Waiting for db relation to be created")
```
ghislainbourgeois, line 209: Not sure this line is required, because when we call the pebble_ready method right after, we will be immediately updating the status based on the relation status.

ghislainbourgeois (changes requested):
I also agree with @Gmerold that we would need a test to cover the behaviour.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
saltiyazan: Yes, that's a good point. I don't know why I added a separate handler :) it's fixed now.

saltiyazan:
> Changes look good to me, but I think we could use some tests for the new behavior.

Added new test cases, that make sure that the charm will go to `blocked` status if the `relation` is missing.
They don't care why is the relation missing. But I guess we trust the framework, and expect the relation to be actually removed and the handler to be called when the `relation_broken` event is emitted.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -112,7 +118,7 @@ def __init__(
         relation_joined_event = getattr(
             self.charm.on, f"{provided_relation_name_with_underscores}_relation_joined"
         )
-        self.framework.observe(pebble_ready_event, self._on_magma_orc8r_pebble_ready)
+        self.framework.observe(pebble_ready_event, self._configure_orc8r)
```
Gmerold: How about `_configure_charm`? Minor detail, but technically we're not configuring the whole Orc8r ;)

Gmerold:
> > Changes look good to me, but I think we could use some tests for the new behavior.
> 
> Added new test cases, that make sure that the charm will go to `blocked` status if the `relation` is missing. They don't care why is the relation missing. But I guess we trust the framework, and expect the relation to be actually removed and the handler to be called when the `relation_broken` event is emitted.

IMO it would be good to add additional tests actually simulating `relation_broken` anyway. It's not the matter of trust we put in the framework, but rather having a safety switch. Such tests would say `this is the behavior we expect in this case` (it's also sort of documentation) and if someone changes this behavior in the future, intentionally or accidentally, he should be informed about it; in this case by a failing test.
What do you think?

saltiyazan:
> > > Changes look good to me, but I think we could use some tests for the new behavior.
> > 
> > 
> > Added new test cases, that make sure that the charm will go to `blocked` status if the `relation` is missing. They don't care why is the relation missing. But I guess we trust the framework, and expect the relation to be actually removed and the handler to be called when the `relation_broken` event is emitted.
> 
> IMO it would be good to add additional tests actually simulating `relation_broken` anyway. It's not the matter of trust we put in the framework, but rather having a safety switch. Such tests would say `this is the behavior we expect in this case` (it's also sort of documentation) and if someone changes this behavior in the future, intentionally or accidentally, he should be informed about it; in this case by a failing test. What do you think?

My confusion came from the fact that the `pebble_ready` handler which is now renamed to `configure_orc8r` will be called, and that method is well covered by unit tests. But having something like `if db_relation is broken the charm should block` makes sense to add, as you said as a safety switch/documentation. :)

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
saltiyazan: Done

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
saltiyazan: It seems like having a separate handler is a better idea, as we can't assume that the relation has been actually completely removed at the time `relation_broken` event is emitted.

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db.py
```diff
@@ -146,6 +146,50 @@ def test_db_relation_added_when_get_status_then_status_is_active(
...
 
+    def test_given_db_relation_not_created_when_pebble_ready_then_status_is_blocked(
+        self,
+    ):
+        self.harness.container_pebble_ready(container_name="magma-orc8r-dummy")
+        assert self.harness.charm.unit.status == BlockedStatus(
+            "Waiting for db relation to be created"
+        )
+
+    @patch("charms.magma_orc8r_libs.v0.orc8r_base_db.Orc8rBase._db_relation_created")
```
Gmerold: We should avoid patching properties, especially private ones. You can just create a db relation instead.

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db.py
```diff
@@ -146,6 +146,50 @@ def test_db_relation_added_when_get_status_then_status_is_active(
...
+    @patch(
+        "charms.magma_orc8r_libs.v0.orc8r_base_db.Orc8rBase.namespace",
+        PropertyMock(return_value="qwerty"),
+    )
+    @patch(
+        "charms.magma_orc8r_libs.v0.orc8r_base_db.Orc8rBase._db_relation_ready",
+        PropertyMock(return_value=True),
+    )
+    @patch(
+        "charms.magma_orc8r_libs.v0.orc8r_base_db.Orc8rBase._get_db_connection_string",
```
Gmerold: As above - I think this test should be adding the db relation and then removing it rather than faking relation availability and emitting `relation_broken` manually.

orchestrator-bundle/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
ghislainbourgeois: Agreed, and we are not calling the other handler after now.

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db.py
```diff
@@ -129,43 +123,70 @@ def test_given_ready_when_get_plan_then_plan_is_filled_with_magma_orc8r_dummy_se
...
+        key_values = {"master": self.TEST_DB_CONNECTION_STRING.__str__()}
+        self.harness.update_relation_data(
+            relation_id=db_relation_id, key_values=key_values, app_or_unit="postgresql-k8s"
+        )
+        self.harness.charm.on.magma_orc8r_dummy_pebble_ready.emit(container)
+        assert self.harness.charm.unit.status == ActiveStatus()
+        self.harness.remove_relation(db_relation_id)
+        assert self.harness.charm.unit.status == BlockedStatus(
+            "Waiting for db relation to be created"
+        )
```
ghislainbourgeois: Small nitpick, but in tests with more than a few lines, I think separating the `given/when/then` sections with a newline helps reading:
```suggestion
        assert self.harness.charm.unit.status == ActiveStatus()
        
        self.harness.remove_relation(db_relation_id)
        
        assert self.harness.charm.unit.status == BlockedStatus(
            "Waiting for db relation to be created"
        )
```

ghislainbourgeois (approved):

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db.py
saltiyazan: Done

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db.py
saltiyazan: Done

orchestrator-bundle/orc8r-libs/tests/test_orc8r_base_db.py
saltiyazan: Done

Gmerold (approved):

PR #110, opened 2022-09-27
Adding provided relation to magma-orc8r-nginx

Gmerold:
# Description

Adds provided relation to `magma-orc8r-nginx`.
This PR is part of [TELCO-184](https://warthogs.atlassian.net/jira/software/c/projects/TELCO/boards/51?selectedIssue=TELCO-184). `magma-orc8r-nginx` needs to provide a relation which then can be consumed by the `magma-orc8r-certifier` which will be responsible for exposing the Orc8r services information to the outside world (in Magma use case that would be AGW).

# Checklist:

### DIFF
--- a/orchestrator-bundle/orc8r-nginx-operator/src/charm.py
+++ b/orchestrator-bundle/orc8r-nginx-operator/src/charm.py
@@ -5,6 +5,7 @@
 """Proxies traffic between nms and obsidian."""
 
 import logging
+import re
 from typing import List, Union
 
 from charms.magma_orc8r_certifier.v0.cert_certifier import CertCertifierRequires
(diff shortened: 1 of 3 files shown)


ghislainbourgeois (approved):

orchestrator-bundle/orc8r-nginx-operator/src/charm.py
```diff
@@ -317,78 +125,87 @@ def _on_magma_orc8r_nginx_pebble_ready(
...
+
+        Updates the status of the `orc8r-nginx` service in the relation data bags so that relation
+        consumers know if it's running or not. If `orc8r-nginx` is not running, event is also
+        deferred.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
+        """
+        self._update_relations()
+        if not self._service_is_running:
```
gruyaume, line 151: Why do we defer the event **after** the action has been made (updating the relations)

Gmerold: If we don't defer, there's no other way to know that the service has started. At the same time, we want to have current service status in the relation data bag, so we set the status first (to whatever it is) and then, if the service is not running yet, we defer to check again later.

gruyaume: ahhh

gruyaume (approved):

PR #111, opened 2022-09-27
Handles db_relation_departed in charms

saltiyazan:
# Description

Please include a summary of the change. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Checklist:

-- using integration test cases


orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
```diff
@@ -102,6 +102,21 @@ async def test_relate_and_wait_for_idle(self, ops_test, setup, build_and_deploy_
...
+    async def test_db_relation_broken_then_joined(self, ops_test, setup, build_and_deploy_charm):
+        rel_id = await ops_test.model.add_relation(
+            relation1=APPLICATION_NAME, relation2="postgresql-k8s:db"
+        )
+        await ops_test.model.add_relation(
+            relation1=APPLICATION_NAME, relation2="orc8r-certifier:cert-admin-operator"
+        )
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
+        await ops_test.model.remove_relation(rel_id)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
```
Gmerold: I think handling `relation_broken` should have a separate test.
WDYT?

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -348,6 +357,18 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+    def _on_database_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Event handler for database relation broken.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
+
+        Returns:
+            None
+        """
+        self.unit.status = BlockedStatus("Waiting for db relation to be created")
```
Gmerold, line 409: Do we need that if we're calling `self._on_magma_nms_magmalte_pebble_ready` next, which will do the same? Perhaps we don't need a separate callback function here (in that case `_on_magma_nms_magmalte_pebble_ready` would probably have to be renamed).

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -652,6 +656,18 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+    def _on_database_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Event handler for database relation broken.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
+
+        Returns:
+            None
+        """
+        self.unit.status = BlockedStatus("Waiting for db relation to be created")
```
Gmerold: As above.

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -64,6 +64,21 @@ async def test_build_and_deploy(self, ops_test, setup, build_and_deploy):
...
+    async def test_db_relation_broken_then_joined(self, ops_test, setup, build_and_deploy_charm):
+        rel_id = await ops_test.model.add_relation(
+            relation1=APPLICATION_NAME, relation2="postgresql-k8s:db"
+        )
+        await ops_test.model.add_relation(
+            relation1=APPLICATION_NAME, relation2="tls-certificates-operator"
+        )
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
+        await ops_test.model.remove_relation(rel_id)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
```
Gmerold: As above.

Gmerold (changes requested):
What about unit tests for these charms?

ghislainbourgeois (commented):
I pretty much have the same comments as @Gmerold

orchestrator-bundle/nms-magmalte-operator/src/charm.py
saltiyazan: Good point indeed, I don't know why I added a separate one.
I fixed it :)

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
saltiyazan: Fixed

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
saltiyazan: > What about unit tests for these charms?

@ghislainbourgeois @Gmerold 
I renamed unit tests not to be testing `pebble_ready` but the functionality of `configuration`.
I don't think we need a separate test that mocks removing a relation. As far as I remember we should trust `ops` and should trust that the handler will be actually called if the event is emitted.
And the functionality (in this case the handler) is tested and covered by many test cases, but by emitting `pebble_ready`

saltiyazan:
> What about unit tests for these charms?

@ghislainbourgeois @Gmerold
I renamed unit tests not to be testing pebble_ready but the functionality of configuration.
I don't think we need a separate test that mocks removing a relation. As far as I remember we should trust ops and should trust that the handler will be actually called if the event is emitted.
And the functionality (in this case the handler) is tested and covered by many test cases, but by emitting pebble_ready

Gmerold:
> > What about unit tests for these charms?
> 
> @ghislainbourgeois @Gmerold I renamed unit tests not to be testing pebble_ready but the functionality of configuration. I don't think we need a separate test that mocks removing a relation. As far as I remember we should trust ops and should trust that the handler will be actually called if the event is emitted. And the functionality (in this case the handler) is tested and covered by many test cases, but by emitting pebble_ready

As already mentioned in the other PR, IMO it would be good to add additional tests actually simulating relation_broken anyway. It's not the matter of trust we put in the framework, but rather having a safety switch. Such tests would say this is the behavior we expect in this case (it's also sort of documentation) and if someone changes this behavior in the future, intentionally or accidentally, he should be informed about it; in this case by a failing test.

saltiyazan:
> > > What about unit tests for these charms?
> > 
> > 
> > @ghislainbourgeois @Gmerold I renamed unit tests not to be testing pebble_ready but the functionality of configuration. I don't think we need a separate test that mocks removing a relation. As far as I remember we should trust ops and should trust that the handler will be actually called if the event is emitted. And the functionality (in this case the handler) is tested and covered by many test cases, but by emitting pebble_ready
> 
> As already mentioned in the other PR, IMO it would be good to add additional tests actually simulating relation_broken anyway. It's not the matter of trust we put in the framework, but rather having a safety switch. Such tests would say this is the behavior we expect in this case (it's also sort of documentation) and if someone changes this behavior in the future, intentionally or accidentally, he should be informed about it; in this case by a failing test.

Unit test cases added.

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
saltiyazan: Seems like it's a good idea to have a separate handler, as we can't assume that the relation has been completely removed at the time `relation_broken` is emitted.

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -267,12 +274,12 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
         self._on_magma_nms_magmalte_pebble_ready(event)
 
     def _on_magma_nms_magmalte_pebble_ready(
-        self, event: Union[PebbleReadyEvent, CertificateAvailableEvent]
+        self, event: Union[PebbleReadyEvent, CertificateAvailableEvent, RelationJoinedEvent]
     ) -> None:
-        """Triggered when pebble is ready.
+        """Configures pebble layer and creates adming user of nms-magmalte.
```
Gmerold:
```suggestion
        """Configures pebble layer and creates admin user of nms-magmalte.
```

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -345,9 +352,20 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
...
             event.database = self.DB_NAME  # type: ignore[attr-defined]
+            self._on_magma_nms_magmalte_pebble_ready(event)
         else:
             event.defer()
 
+    def _on_database_relation_broken(self, event: RelationBrokenEvent):
+        """Event handler for database relation broken.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
```
Gmerold:
```suggestion
            event (RelationBrokenEvent): Juju event
```

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -267,12 +274,12 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
     def _on_magma_nms_magmalte_pebble_ready(
-        self, event: Union[PebbleReadyEvent, CertificateAvailableEvent]
+        self, event: Union[PebbleReadyEvent, CertificateAvailableEvent, RelationJoinedEvent]
     ) -> None:
-        """Triggered when pebble is ready.
+        """Configures pebble layer and creates adming user of nms-magmalte.
 
         Args:
-            event (PebbleReadyEvent, CertificateAvailableEvent): Juju event
+            event (PebbleReadyEvent, CertificateAvailableEvent, RelationBrokenEvent): Juju event
```
Gmerold:
```suggestion
            event (PebbleReadyEvent, CertificateAvailableEvent, RelationJoinedEvent): Juju event
```

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -651,6 +655,19 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
...
             event.database = self.DB_NAME  # type: ignore[attr-defined]
+            self._on_magma_orc8r_certifier_pebble_ready(event)
+        else:
+            event.defer()
+
+    def _on_database_relation_broken(self, event: RelationBrokenEvent):
+        """Event handler for database relation broken.
+
+        Args:
+            event (RelationJoinedEvent): Juju event
```
Gmerold:
```suggestion
            event (RelationBrokenEvent): Juju event
```

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -661,7 +663,40 @@ def test_given_certificates_relation_not_created_when_on_pebble_ready_then_statu
...
+            relation_id=certificates_relation_id, remote_unit_name="vault-k8s/0"
+        )
+        key_values = {"master": self.TEST_DB_CONNECTION_STRING.__str__()}
+        self.harness.update_relation_data(
+            relation_id=db_relation_id, app_or_unit="postgresql-k8s", key_values=key_values
+        )
+
+        self.harness.container_pebble_ready(container_name="magma-orc8r-certifier")
+        assert self.harness.charm.unit.status == ActiveStatus()
+        self.harness.charm.on.db_relation_broken.emit(self.harness.model.get_relation("db"))
```
Gmerold: Why not just remove the relation to trigger the event emission?

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
```diff
@@ -209,6 +209,39 @@ def test_given_relations_are_created_and_certs_are_stored_when_pebble_ready_then
...
+            "user=test_db_user"
+        }
+        self.harness.update_relation_data(
+            relation_id=db_relation_id, key_values=key_values, app_or_unit="postgresql-k8s"
+        )
+        self.harness.container_pebble_ready(container_name="magma-nms-magmalte")
+
+        self.assertEqual(self.harness.charm.unit.status, ActiveStatus())
+
+        self.harness.charm.on.db_relation_broken.emit(self.harness.model.get_relation("db"))
```
Gmerold: As below - why not remove the relation instead of emitting that manually?

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
```diff
@@ -102,6 +103,15 @@ async def test_relate_and_wait_for_idle(self, ops_test, setup, build_and_deploy_
         )
         await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
 
+    async def test_remove_db_application(self, ops_test, setup, build_and_deploy_charm):
+        await ops_test.model.applications[DB_APPLICATION_NAME].remove(force=True)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
+        await self._deploy_postgresql(ops_test)
```
Gmerold, line 141: How about having this a separate test?

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -64,6 +65,15 @@ async def test_build_and_deploy(self, ops_test, setup, build_and_deploy):
         )
         await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=1000)
 
+    async def test_remove_db_application(self, ops_test, setup, build_and_deploy):
+        await ops_test.model.applications[DB_APPLICATION_NAME].remove(force=True)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="blocked", timeout=1000)
+        await ops_test.model.deploy("postgresql-k8s", application_name=DB_APPLICATION_NAME)
```
Gmerold: Separate test?

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
saltiyazan: Done

orchestrator-bundle/nms-magmalte-operator/tests/unit/test_charm.py
saltiyazan: Done

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
saltiyazan: Done, testing only for blocked status

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
saltiyazan: Done, testing only for blocked status

orchestrator-bundle/nms-magmalte-operator/src/charm.py
```diff
@@ -460,6 +480,10 @@ def _create_nms_admin_user(self, email: str, password: str, organization: str) -
         Returns:
             None
         """
+        if not self._service_is_running:
```
Gmerold, line 528: Shouldn't we raise an error here or something? I mean if the service is not running, this method will return, `_on_magma_nms_magmalte_pebble_ready` will think everything went smoothly and `ActiveStatus` will be set. Or am I missing something?

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
```diff
@@ -109,6 +110,10 @@ async def test_scale_up(self, ops_test, setup, build_and_deploy_charm):
             apps=[APPLICATION_NAME], status="active", timeout=1000, wait_for_exact_units=2
         )
 
+    async def test_remove_db_application(self, ops_test, setup, build_and_deploy_charm):
```
Gmerold: Don't we want to test re-deployment of the DB to make sure charm gets back to `Active` status?

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -71,6 +72,10 @@ async def test_build_and_deploy_and_scale_up(self, ops_test, setup, build_and_de
             apps=[APPLICATION_NAME], status="active", timeout=1000, wait_for_exact_units=2
         )
 
+    async def test_remove_db_application(self, ops_test, setup, build_and_deploy):
```
Gmerold, line 76: As above.

orchestrator-bundle/nms-magmalte-operator/src/charm.py
saltiyazan: Correct, fixed :)

orchestrator-bundle/nms-magmalte-operator/tests/integration/test_integration.py
saltiyazan: Fixed

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
saltiyazan: Fixed

orchestrator-bundle/nms-magmalte-operator/src/charm.py
Gmerold: Cool, but now it also needs to be handled somewhere, right?
BTW, perhaps having a custom exception would make sense? I guess we don't want our handler to catch any exception.

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -662,7 +664,40 @@ def test_given_certificates_relation_not_created_when_on_pebble_ready_then_statu
...
+        self.harness.add_relation_unit(
+            relation_id=certificates_relation_id, remote_unit_name="vault-k8s/0"
+        )
+        key_values = {"master": self.TEST_DB_CONNECTION_STRING.__str__()}
+        self.harness.update_relation_data(
+            relation_id=db_relation_id, app_or_unit="postgresql-k8s", key_values=key_values
+        )
+
+        self.harness.container_pebble_ready(container_name="magma-orc8r-certifier")
+        assert self.harness.charm.unit.status == ActiveStatus()
```
gruyaume: We shouldn't have unrelated asserts in a unit test, there should probably be two different tests that we validate the behaviour for.

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
```diff
@@ -14,13 +14,14 @@
...
+DB_APPLICATION_NAME = "postgresql-k8s"
 
 
 class TestOrc8rCertifier:
     @pytest.fixture(scope="module")
     @pytest.mark.abort_on_fail
     async def setup(self, ops_test):
-        await ops_test.model.deploy("postgresql-k8s", application_name="postgresql-k8s")
+        await ops_test.model.deploy("postgresql-k8s", application_name=DB_APPLICATION_NAME)
         await ops_test.model.wait_for_idle(apps=["postgresql-k8s"], status="active", timeout=1000)
```
gruyaume, line 25: No need to wait for idle for the postgresql-k8s charm. In reality we don't wait for it to be active before deploying more apps

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
saltiyazan: The goal of this is not to validate that it goes to `active` status, but to show that it will go to `blocked` after having been in `active` if a required dependency is removed. But I'm okay with removing this assert if that is cleaner.

gruyaume: I understand, we don't need the first assert in this case

orchestrator-bundle/orc8r-certifier-operator/tests/integration/test_integration.py
saltiyazan: Sure, makes sense, however I suggest to address it in a new PR :)

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
saltiyazan: removed :)

orchestrator-bundle/nms-magmalte-operator/src/charm.py
saltiyazan: Done

orchestrator-bundle/nms-magmalte-operator/src/charm.py
saltiyazan: Done

orchestrator-bundle/orc8r-certifier-operator/tests/unit/test_charm.py
saltiyazan: Done

Gmerold (approved):
Good job!

gruyaume (approved):

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -283,6 +287,20 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
         """
         if self.unit.is_leader():
             event.database = self.DB_NAME  # type: ignore[attr-defined]
+            self._on_magma_orc8r_certifier_pebble_ready(event)
+        else:
+            event.defer()
```
ghislainbourgeois:
```suggestion
```

If the unit is not leader, we should just ignore the event, if not, this will create a defer loop.

ghislainbourgeois (changes requested):

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
saltiyazan: Correct, fixed.

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
```diff
@@ -281,8 +285,21 @@ def _on_database_relation_joined(self, event: RelationJoinedEvent) -> None:
         Returns:
             None
         """
-        if self.unit.is_leader():
-            event.database = self.DB_NAME  # type: ignore[attr-defined]
+        if not self.unit.is_leader():
+            return
+        event.database = self.DB_NAME  # type: ignore[attr-defined]
+        self._on_magma_orc8r_certifier_pebble_ready(event)
```
ghislainbourgeois, line 291: This change will prevent the non-leader units from having the pebble layer configured, is this what we really want?

ghislainbourgeois (changes requested):

orchestrator-bundle/orc8r-certifier-operator/src/charm.py
saltiyazan: So far we would allow only leader units to set `self.DB_NAME` which will block pebble layer configuration if not set, causing pebble not be configured by non-leader units.
So there is no change from that aspect here.
Do you propose to revisit the logic of this part?

ghislainbourgeois: No, I think this is ok for now, but this basically mean that having more than one unit for this charm is pointless. It is probably something we want to improve at a later time.

ghislainbourgeois (approved):

PR #113, opened 2022-09-28
Dev handle relation broken events

saltiyazan:
# Description

Please include a summary of the change. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Checklist:


orchestrator-bundle/orc8r-metricsd-operator/tests/unit/test_charm.py
```diff
@@ -55,6 +55,66 @@ def test_given_no_relations_created_when_pebble_ready_then_charm_goes_to_blocked
...
+
+        self.harness.remove_relation(
+            self.harness.model.get_relation(
+                "alertmanager-configurer-k8s"
+            ).id  # type: ignore[union-attr]
+        )
+
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus("Waiting for relation(s) to be created: alertmanager-configurer-k8s"),
```
Gmerold, line 91: At this point we're already missing two relations, but the status only reports one. Do you think it would make sense to check for all missing relations?
PS. I know that once this missing relation joins, charm will detect and report any other missing relations.

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -255,6 +255,36 @@ def test_given_magma_orc8r_obsidian_relation_not_created_when_pebble_ready_event
...
+        )
+
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus("Waiting for relation(s) to be created: magma-orc8r-obsidian"),
+        )
+
+        self.harness.remove_relation(
+            self.harness.model.get_relation(
+                "magma-orc8r-bootstrapper"
```
Gmerold, line 286: As above.

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
```diff
@@ -512,11 +449,62 @@ def test_given_metrics_endpoint_relation_is_not_created_when_pebble_ready_then_s
...
+        self.harness.remove_relation(
+            self.harness.model.get_relation("metrics-endpoint").id  # type: ignore[union-attr]
+        )
+
+        self.assertEqual(
+            self.harness.charm.unit.status,
+            BlockedStatus("Waiting for relation(s) to be created: metrics-endpoint"),
+        )
+
+        self.harness.remove_relation(
```
Gmerold, line 442: As above.

orchestrator-bundle/orc8r-metricsd-operator/tests/unit/test_charm.py
saltiyazan: Do you think having separate test cases is a better idea? for me it seems too repetitive, and it won't test any other behaviour than this one.  I mean here we can see that the charm is responding as its status stays blocked with new message :)
Another option would be to have one test with one relation, but then which relation gets chosen, and of course that is more loose.
Let me know your thoughts on which approach would be the best?

Gmerold: I was thinking more about the implementation rather than the test. Right now `on_relation_broken` sets the status for the relation that triggered the event. My idea was to actually iterate over the relations to find out the missing ones (just like we do in `configure_XXX_charm`.

saltiyazan: There is one problem with that: We are not sure that the relation is actually removed after it is broken :(
But we can call `_relation_created` to check if any other relations are still missing, but in that case we might miss the relation that is actually broken as `relation_created` might return `true` for that.
What do you think?

Gmerold: Oh yeah, there's that... true. Forget my comment then.

Gmerold (commented):
It looks good to me. I'll wait with the approval for green itests tho (just in case there's some more changes).

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
```diff
@@ -567,3 +558,42 @@ def test_given_pebble_ready_when_certificate_available_then(self, patch_push):
                 call(path="/var/opt/magma/certs/admin_operator.key.pem", source=private_key),
             ]
         )
+
+    def _create_relations(self):
```
gruyaume: I am not sure about this method, its name suggests that it creates a relation but the content is much more complex as it contains multiple units, relation data on 2 sides. I think it should  take some input parameters and have some docstring.

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
```diff
@@ -487,13 +479,13 @@ def _set_log_verbosity_action(self, event: ActionEvent) -> None:
             for line in e.stderr.splitlines():
                 logger.error("    %s", line)
 
-    def _on_magma_orc8r_orchestrator_pebble_ready(  # noqa: C901
-        self, event: Union[PebbleReadyEvent, CertificateAvailableEvent]
+    def _configure_magma_orc8r_orchestrator(  # noqa: C901
```
ghislainbourgeois: Since you simplified this method a lot, I think we can remove `# noqa: C901`

ghislainbourgeois (changes requested):

orchestrator-bundle/orc8r-orchestrator-operator/tests/unit/test_charm.py
saltiyazan: I changed it, let me know what do you think :)

orchestrator-bundle/orc8r-orchestrator-operator/src/charm.py
saltiyazan: Removed

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -478,6 +408,35 @@ def test_given_metricsd_service_running_when_metricsd_relation_joined_then_servi
             {"active": "True"},
         )
 
+    def _create_active_relation(self, relation_name: str, remote_app: str):
+        """Creates a relation between orc8r-nginx and a remote app and activates it.
+
+        Args:
+            relation_name: str
```
gruyaume:
```suggestion
            relation_name (str): Relation name
```

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -478,6 +408,35 @@ def test_given_metricsd_service_running_when_metricsd_relation_joined_then_servi
...
+            remote_app: str
```
gruyaume:
```suggestion
            remote_app (str): Remote application
```

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -478,6 +408,35 @@ def test_given_metricsd_service_running_when_metricsd_relation_joined_then_servi
             {"active": "True"},
         )
 
+    def _create_active_relation(self, relation_name: str, remote_app: str):
+        """Creates a relation between orc8r-nginx and a remote app and activates it.
```
gruyaume: There is no such thing as activating a relation. Here we add a given key value that internally for Magma projects is used to send the information that the service running in the workload of the remote app is running.

orchestrator-bundle/orc8r-metricsd-operator/tests/unit/test_charm.py
```diff
@@ -55,6 +55,66 @@ def test_given_no_relations_created_when_pebble_ready_then_charm_goes_to_blocked
...
+        )
+
+        self.harness.remove_relation(
+            self.harness.model.get_relation("prometheus-k8s").id  # type: ignore[union-attr]
+        )
+
+        self.assertEqual(
```
gruyaume, line 106: We have too many assert in a given unit test, unit tests should have 1 reason to fail, not multiple. It's fine to have multiple asserts if they all validate 1 behaviour but we shouldn't chain actions and asserts in 1 test.

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
```diff
@@ -157,6 +174,11 @@ def _on_magma_orc8r_metricsd_relation_joined(self, event: RelationJoinedEvent) -
             event.defer()
             return
 
+    def _on_required_relation_broken(self, event: RelationBrokenEvent):
```
gruyaume, line 177: Please add docstring

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
saltiyazan: Fixed

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
saltiyazan: Fixed

orchestrator-bundle/orc8r-nginx-operator/tests/unit/test_charm.py
saltiyazan: Fixed

orchestrator-bundle/orc8r-metricsd-operator/tests/unit/test_charm.py
saltiyazan: The goal here was to avoid having separate test cases for the same thing (I felt it would be very repetitive)
However I moved them to separate cases now.

orchestrator-bundle/orc8r-metricsd-operator/src/charm.py
saltiyazan: done

Gmerold (approved):

gruyaume (approved):

ghislainbourgeois (approved):

PR #114, opened 2022-09-28
Adds integration tests

gruyaume:
# Description

- Adds integration tests to bundle:
  - Install Microk8s with metallb add-on
  - Packs all orchestrator charms
  - Renders and deploys local bundle
  - Waits for all charms to be Active
  - Retrieves pfx package and password
  - Retrieves load-balancer services
  - Makes an API command to orc8r using the provided certificate

- Improves the bundle rendering script
  -  Include option for local deployment
  - Adds unit testing

# Checklist:

- [ ] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [ ] I have performed a self-review of my own code
- [ ] I validated that new and existing unit tests pass locally with my changes


orchestrator-bundle/orc8r-bundle/tests/integration/test_integration.py
```diff
@@ -2,31 +2,222 @@
...
+
+    """
+    orc8r_orchestrator_unit = ops_test.model.units["orc8r-orchestrator/0"]  # type: ignore[union-attr]  # noqa: E501
+    load_balancer_action = await orc8r_orchestrator_unit.run_action(
+        action_name="get-load-balancer-services"
+    )
+    load_balancer_action_output = await ops_test.model.get_action_output(  # type: ignore[union-attr]  # noqa: E501
+        action_uuid=load_balancer_action.entity_id, wait=60
+    )
+    return (
```
saltiyazan, line 73: The order matters here, as the caller of this function will expect 
```(
            orc8r_bootstrap_nginx_ip,
            orc8r_clientcert_nginx_ip,
            orc8r_nginx_proxy_ip,
            nginx_proxy_ip,
        )
```
Could you expose this information? for example in the docstring?

orchestrator-bundle/orc8r-bundle/tests/integration/orchestrator.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+            pem_file.write(certificate.public_bytes(Encoding.PEM))
+        yield t_pem.name
+
+
+class Orc8r:
+    """Class that represents a Magma Orchestrator Instance"""
+
+    def __init__(self, url: str, admin_operator_pfx_path: str, admin_operator_pfx_password: str):
+        self.url = url
```
saltiyazan, line 53: might be an overkill, but shouldn't we validate the url here?

orchestrator-bundle/orc8r-bundle/tests/integration/test_integration.py
gruyaume: Fixed

orchestrator-bundle/orc8r-bundle/tests/integration/orchestrator.py
gruyaume: Humm sounds overkill a bit and the idea is to use a lib specifically for communicating with orc8r somewhat soon

saltiyazan: yeah I thought so too, my concern is that if `url=f"https://{orc8r_nginx_proxy_ip}/magma/v1/"` is changed to `url=f"https://{orc8r_nginx_proxy_ip}/magma/v1"` (/ is removed from the end for example) in `test_integration.py` the test would raise an exception.

saltiyazan (approved):

orchestrator-bundle/orc8r-bundle/tests/integration/orchestrator.py
gruyaume: Yes it would be ultimately, and that's probably something we should consider in the `magma-orc8r-client` library.

PR #137, opened 2022-10-21
Adds series to bootstrapper integration tests

gruyaume:
# Description

Adds series to bootstrapper integration tests

# Checklist:

- [ ] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [ ] I have performed a self-review of my own code
- [ ] I validated that new and existing unit tests pass locally with my changes

### DIFF
--- a/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
+++ b/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
@@ -80,8 +80,8 @@ class Orc8rBase(Object):
         self,
         charm: CharmBase,
         startup_command: str,
-        required_relations: list = None,
-        additional_environment_variables: dict = None,
+        required_relations: list = [],
+        additional_environment_variables: dict = {},
     ):
         """Observes common events for all Orchestrator charms."""
         super().__init__(charm, "orc8r-base")
--- a/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
+++ b/orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
@@ -80,8 +80,8 @@ class Orc8rBase(Object):
         self,
         charm: CharmBase,
         startup_command: str,
-        required_relations: list = [],
-        additional_environment_variables: dict = {},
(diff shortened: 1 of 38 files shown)


ghislainbourgeois (approved):

orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
```diff
@@ -80,8 +80,8 @@ def __init__(
         self,
         charm: CharmBase,
         startup_command: str,
-        required_relations: list = None,
-        additional_environment_variables: dict = None,
+        required_relations: list = [],
+        additional_environment_variables: dict = {},
```
ghislainbourgeois: This is dangerous, in most cases, because the default list or dict that is created will be the same object for all instances. In this case, it will probably not cause any issues, but I would change it anyway to get rid of the dangerous construct.

The best way to deal with default objects is to have the default set to `None` and immediately check it:

```
def f(my_list: list = None):
    if my_list is None:
        my_list = []
```

orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
```diff
@@ -101,7 +101,7 @@ def __init__(
         self,
         charm: CharmBase,
         startup_command: str,
-        additional_environment_variables: dict = None,
+        additional_environment_variables: dict = {},
```
ghislainbourgeois: As above

ghislainbourgeois (changes requested):

orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base.py
saltiyazan: Fixed :)

orc8r-libs/lib/charms/magma_orc8r_libs/v0/orc8r_base_db.py
saltiyazan: Fixed

ghislainbourgeois (approved):

PR #139, opened 2022-10-25
Adds an option to promote all charms

saltiyazan:
# Description

Adds an `all` option to promote all of the charms.

# Checklist:

### DIFF
--- a/.github/workflows/charms.promote.yaml
+++ b/.github/workflows/charms.promote.yaml
@@ -49,8 +49,9 @@ on:
           - orc8r-tenants
 
 jobs:
-  promote:
-    name: Promote Charm
+  promote-all:
+    if: github.event.inputs.charm-name == 'all'
+    name: Promote All Charms
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/charms.promote.yaml
```diff
@@ -71,9 +106,20 @@ jobs:
       - name: Promote Charm
         uses: canonical/charming-actions/release-charm@1.0.3
         with:
+          if: github.event.inputs.charm-name != 'all'
```
ghislainbourgeois: This `if` is not at the same indentation level as the other block, can you validate which is right?

ghislainbourgeois (changes requested):

.github/workflows/charms.promote.yaml
saltiyazan: Fixed. I changed the whole logic of condition checks, it wasn't working as it was.

ghislainbourgeois (approved):

Gmerold (approved):

PR #141, opened 2022-10-28
Telco 344 fixing unable to relate multiple agws to one orc8r

Gmerold:
# Description

Updates the `magma-orchestrator-interface` library in `orc8r-nginx-operator` to fix the bug preventing relating multiple AGWs to one Orc8r.
Additionally removes unnecessary parts of itests setup.

# Checklist:


bundle-local.yaml
```diff
@@ -0,0 +1,302 @@
+bundle: kubernetes
```
gruyaume: I don't think we want this. The existing bundle template can be used for generating the local bundle.

gruyaume (changes requested):

bundle-local.yaml
Gmerold: Oh, didn't know that.
Removed.

gruyaume (approved):

saltiyazan (approved):

PR #144, opened 2022-12-02
Adding fluentd charm

Gmerold:
# Description

Adds fluentd charm.
**NOTE:**
This is just an initial PR and this charm is **not ready** to be added to the bundle yet. In current shape, it won't even start, because fluentd certificates are still missing. Fluentd certs are handles in [this story](https://warthogs.atlassian.net/browse/TELCO-352).

# Checklist:


gruyaume:
I don't think the charm should be part of the charmed magma repo, we should leverage the existing [project](https://github.com/canonical/fluentd-elasticsearch-operator) and make the changes there.

fluentd/src/config_files/output.conf.j2
```diff
@@ -0,0 +1,58 @@
+<match eventd>
```
gruyaume: There are 2 blocks here, do we know if they are both used in Magma? Do we see logs tagged with `eventd` and with `magma`

fluentd/src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+    @property
+    def _domain_config(self) -> Optional[str]:
+        """Returns domain config.
+
+        Returns:
+            str: Domain
+        """
+        return self.model.config.get("domain")
+
+    def _cert_is_stored(self, cert_path: str) -> bool:
```
gruyaume: The method name doesn't match with what it does. Here we could pass it anything and it would validate that it exists on the container (not necessariliy a cert)

fluentd/src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+from ops.pebble import Layer
+
+logger = logging.getLogger(__name__)
+
+
+class FluentdElasticsearchCharm(CharmBase):
+
+    BASE_CERTS_PATH = "/certs"
+    CONFIG_DIRECTORY = "/etc/fluent/config.d"
+    REQUIRED_RELATIONS = ["cert-certifier"]
```
gruyaume: Fluentd requires a relation to anything that gives it certificates, not specifically certifier.

fluentd/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Config
+
+### Config options
+
+- **domain**: Fluentd domain (same as Orc8r domain).
+- **elasticsearch-url**: ElasticSearch URL (example: orc8r-elasticsearch:9200).
+- **elasticsearch-schema**: "http" or "https" depending on whether the ElasticSearch endpoint 
+supports SSL. Defaults to https.
+- **elasticsearch-ssl-version**: SSL version. Defaults to TLSv1.
```
gruyaume: That doesn't need to be a config, we can just use the most recent in the charm

fluentd/README.md
```diff
@@ -0,0 +1,46 @@
...
+--config elasticsearch-url="yourelasticsearch:9200" 
+```
+
+## Config
+
+### Config options
+
+- **domain**: Fluentd domain (same as Orc8r domain).
+- **elasticsearch-url**: ElasticSearch URL (example: orc8r-elasticsearch:9200).
+- **elasticsearch-schema**: "http" or "https" depending on whether the ElasticSearch endpoint
```
gruyaume: This should be automatically generated if we leverage the tls certificates relation

fluentd/README.md
```diff
@@ -0,0 +1,46 @@
...
+juju deploy fluentd --trust --channel edge \
+--config domain="example.com" \
+--config elasticsearch-url="yourelasticsearch:9200" 
+```
+
+## Config
+
+### Config options
+
+- **domain**: Fluentd domain (same as Orc8r domain).
```
gruyaume:
```suggestion
- **domain**: Fluentd domain
```

fluentd/src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+        super().__init__(*args)
+        self._container_name = self._service_name = "fluentd"
+        self._container = self.unit.get_container(self._container_name)
+        self._service_patcher = KubernetesServicePatch(
+            charm=self,
+            ports=[ServicePort(name="forward", port=24224)],
+            service_type="LoadBalancer",
+            service_name="fluentd",
+            additional_annotations={
+                "external-dns.alpha.kubernetes.io/hostname": f"fluentd.{self._domain_config}"
```
gruyaume: Is this really needed?

fluentd/src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+        Returns:
+            str: Rendered config file's content
+        """
+        file_loader = FileSystemLoader(config_templates_dir)
+        env = Environment(loader=file_loader)
+        template = env.get_template(template_file_name)
+        elasticsearch_config = self._get_elasticsearch_config()
+        fluentd_config = self._get_fluentd_config()
+        return template.render(
+            domain=self._domain_config,
```
gruyaume: I don't think `domain` is used in the template

fluentd/README.md
```diff
@@ -0,0 +1,46 @@
...
+- **elasticsearch-url**: ElasticSearch URL (example: orc8r-elasticsearch:9200).
+- **elasticsearch-schema**: "http" or "https" depending on whether the ElasticSearch endpoint 
+supports SSL. Defaults to https.
+- **elasticsearch-ssl-version**: SSL version. Defaults to TLSv1.
+- **fluentd-chunk-limit-size**: The size limit of the received chunk. If the chunk size is larger
```
gruyaume: I'd remove the config for this and keep the charm minimal

fluentd/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Config options
+
+- **domain**: Fluentd domain (same as Orc8r domain).
+- **elasticsearch-url**: ElasticSearch URL (example: orc8r-elasticsearch:9200).
+- **elasticsearch-schema**: "http" or "https" depending on whether the ElasticSearch endpoint 
+supports SSL. Defaults to https.
+- **elasticsearch-ssl-version**: SSL version. Defaults to TLSv1.
+- **fluentd-chunk-limit-size**: The size limit of the received chunk. If the chunk size is larger 
+than this value, the received chunk is dropped. Defaults to 2M.
+- **fluentd-queue-limit-length**: The limit of the chunk queue length. If the chunk queue length
```
gruyaume: I'd remove the config for this and keep the charm minimal

fluentd/src/charm.py
```diff
@@ -0,0 +1,327 @@
...
+from ops.main import main
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+from ops.pebble import Layer
+
+logger = logging.getLogger(__name__)
+
+
+class FluentdElasticsearchCharm(CharmBase):
+
+    BASE_CERTS_PATH = "/certs"
```
gruyaume: How does fluentd knows to use those certs? I see nothing pointing to this path

fluentd/README.md
```diff
@@ -0,0 +1,46 @@
+# fluentd
+
```
gruyaume: I'd add a warning to tell that this charm is hardcoded to work with Magma only.

fluentd/charmcraft.yaml
```diff
@@ -0,0 +1,13 @@
+type: "charm"
+bases:
+  - build-on:
+      - name: "ubuntu"
+        channel: "20.04"
+    run-on:
+      - name: "ubuntu"
+        channel: "20.04"
+
+parts:
```
gruyaume: Is this part necessary?

gruyaume (changes requested):

fluentd/README.md
Gmerold: The last line of the description already says `This fluentd charm 
is specifically built to forward Magma logs to elasticsearch.`
Do you think it should be more exposed? I mean this charm is under `charmed-magma-orchestrator` repo anyway.

fluentd/charmcraft.yaml
Gmerold: No, I forgot to remove it after changing `ops` not be fetched from git.
Removed.

fluentd/src/charm.py
Gmerold: Once certs are in place, `forward-input.conf` will tell fluentd to use this path:
```
<source>
  @type forward
  port 24224
  bind 0.0.0.0
  <transport tls>
    ca_path /certs/certifier.pem
    cert_path /certs/fluentd.pem
    private_key_path /certs/fluentd.key
    client_cert_auth true
  </transport>
</source>
```

fluentd/src/charm.py
Gmerold: Ups, my bad.
Removed.

fluentd/src/charm.py
Gmerold: No, not really.

fluentd/README.md
Gmerold: `domain` removed completely.

fluentd/src/charm.py
Gmerold: Updated the name.

fluentd/src/config_files/output.conf.j2
Gmerold: Not sure. Didn't get that far with Magma usage ;) I'd assume it is needed, because that's extracted directly from the official deployment.

fluentd/src/charm.py
Gmerold: If fact we need certifier cert specifically. That's because when AGW pushes logs to fluentd it uses cert signed by the `certifier.pem` and fluentd needs to validate that cert.
Anything providing certs is true for fluentd cert only.

fluentd/README.md
gruyaume: Yeah if we keep it under charmed-magma I don'y really care so you can ignore my messge.

fluentd/src/charm.py
Gmerold: Certs related stuff removed from this PR. It will be addressed in another PR.

fluentd/README.md
Gmerold: Certs related stuff removed from this PR. It will be addressed in another PR.

fluentd/README.md
Gmerold: As discussed this can be useful in case of bigger deployments.

fluentd/README.md
Gmerold: As above.

fluentd/README.md
Gmerold: Certs related stuff removed from this PR. It will be addressed in another PR.

fluentd/README.md
```diff
@@ -0,0 +1,35 @@
+# fluentd
```
gruyaume:
```suggestion
# fluentd-elasticsearch
```

fluentd/metadata.yaml
```diff
@@ -0,0 +1,44 @@
...
+  is specifically built to forward Magma logs to elasticsearch.
+website: https://github.com/canonical/charmed-magma-orchestrator
+
+containers:
+  fluentd:
+    resource: fluentd-elasticsearch-image
+    mounts:
+      - storage: buffer
+        location: /var/log/fluentd-buffers
+      - storage: certs
```
gruyaume: I'd suggest removing certs and re-adding that in the PR that adds certs

fluentd/metadata.yaml
```diff
@@ -0,0 +1,44 @@
...
+        location: /certs
+      - storage: config
+        location: /etc/fluent/config.d
+
+storage:
+  buffer:
+    type: filesystem
+    description: Fluentd buffer
+    minimum-size: 10M
+  certs:
```
gruyaume: I'd suggest removing certs and re-adding that in the PR that adds certs

fluentd/src/charm.py
```diff
@@ -0,0 +1,208 @@
...
+        self._service_patcher = KubernetesServicePatch(
+            charm=self,
+            ports=[ServicePort(name="forward", port=24224)],
+            service_type="LoadBalancer",
+            service_name="fluentd",
+        )
+
+        self.framework.observe(self.on.fluentd_pebble_ready, self._configure)
+        self.framework.observe(self.on.config_changed, self._configure)
+
```
gruyaume: All the config files that are static can be written to on install while the dynamic ones can be written to on config_changed. And like this we can also drop the `pebble_ready` handler and move it all to the config_changed.

fluentd/src/charm.py
```diff
@@ -0,0 +1,208 @@
...
+    def __init__(self, *args):
+        """An instance of this object everytime an event occurs."""
+        super().__init__(*args)
+        self._container_name = self._service_name = "fluentd"
+        self._container = self.unit.get_container(self._container_name)
+        self._service_patcher = KubernetesServicePatch(
+            charm=self,
+            ports=[ServicePort(name="forward", port=24224)],
+            service_type="LoadBalancer",
+            service_name="fluentd",
```
gruyaume:
```suggestion
```

fluentd/src/charm.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def _get_elasticsearch_config(self) -> dict:
+        """Extracts ElasticSearch configuration from Juju config.
+
+        Returns:
+            dict: ElasticSearch configuration
+        """
+        # TODO: Elasticsearch url and port should be passed through relationship
+        elasticsearch_url = self.model.config.get("elasticsearch-url")
+        elasticsearch_url_split = elasticsearch_url.split(":")  # type: ignore[union-attr]
```
gruyaume: You can get rid of the `# type: ignore[union-attr]` if you handle the case where there is no elasticsearch_url set

fluentd/src/charm.py
```diff
@@ -0,0 +1,208 @@
...
+        return {
+            "host": elasticsearch_url_split[0],
+            "port": elasticsearch_url_split[1],
+        }
+
+    def _get_fluentd_config(self) -> dict:
```
gruyaume: I'm not sure of the value of combining the 2 together in 1 get, I thing having properties for the 2 values is just fine

gruyaume:
```python
@property
def _config_fluentd_chunk_limit_size:
    return self.model.config.get("fluentd-chunk-limit-size")
```

fluentd/src/charm.py
```diff
@@ -0,0 +1,208 @@
...
+            dict: Fluentd configuration
+        """
+        return {
+            "chunk_limit_size": self.model.config.get("fluentd-chunk-limit-size"),
+            "queue_limit_length": self.model.config.get("fluentd-queue-limit-length"),
+        }
+
+    @property
+    def _elasticsearch_url_is_valid(self) -> bool:
+        """Checks whether given Elasticsearch URL is valid or not.
```
gruyaume:
```suggestion
        """Returns whether given Elasticsearch URL is valid or not.
```

fluentd/README.md
Gmerold: I wanted this to match the name of the charm; and charm name can't be `fluentd-elasticsearch`, because it's already taken.

fluentd/metadata.yaml
Gmerold: Sure.

fluentd/metadata.yaml
Gmerold: Sure.

fluentd/src/charm.py
Gmerold: Done.

fluentd/README.md
gruyaume: Well it's taken by us. We can replace fluentd with fluentd-elasticsearch and this instance here becomes this charm.

Gmerold: Renamed. Old repo archived.

ghislainbourgeois (approved):

gruyaume (approved):

PR #150, opened 2023-01-12
Set ActiveStatus after successful certificate renewal

ghislainbourgeois:
# Description

Set ActiveStatus after successful certificate renewal.

# Checklist:

### DIFF
--- a/orc8r-certifier-operator/tests/unit/test_charm.py
+++ b/orc8r-certifier-operator/tests/unit/test_charm.py
@@ -18,7 +18,7 @@ from cryptography import x509
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.serialization import pkcs12
 from ops import testing
-from ops.model import BlockedStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
 from ops.pebble import PathError
 from pgconnstr import ConnectionString  # type: ignore[import]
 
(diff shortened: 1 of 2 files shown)


orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -791,6 +791,68 @@ def test_given_relations_are_created_and_certificates_are_stored_when_pebble_rea
...
+                        "SQL_DIALECT": "psql",
+                        "SERVICE_HOSTNAME": "magma-orc8r-certifier",
+                        "SERVICE_REGISTRY_MODE": "k8s",
+                        "SERVICE_REGISTRY_NAMESPACE": self.model_name,
+                    },
+                }
+            },
+        }
+        updated_plan = self.harness.get_container_pebble_plan("magma-orc8r-certifier").to_dict()
+        self.assertEqual(expected_plan, updated_plan)
```
gruyaume: The test title mentions `pebble_plan_already_in_place` as part of the initial state. I'm not sure we should assert it as part of the expected behavior. I think we're testing 2 behaviors in 1 test here.

ghislainbourgeois: Yes, the assert is basically there to make sure that the pebble plan is in fact already in place. I can remove it, it is not doing much.

gruyaume: Well if we think the validation is useful, it should have its own test. If not I'd remove it yes.

ghislainbourgeois: Removed

orc8r-certifier-operator/tests/unit/test_charm.py
```diff
@@ -791,6 +791,37 @@ def test_given_relations_are_created_and_certificates_are_stored_when_pebble_rea
...
+            relation_id=certificates_relation_id, remote_unit_name="vault-k8s/0"
+        )
+        key_values = {"master": self.TEST_DB_CONNECTION_STRING.__str__()}
+        self.harness.update_relation_data(
+            relation_id=db_relation_id, app_or_unit="postgresql-k8s", key_values=key_values
+        )
+
+        self.harness.container_pebble_ready(container_name="magma-orc8r-certifier")
+
+        self.harness.charm.unit.status = WaitingStatus("Waiting for certs")
```
Gmerold: How about explicitly adding a pebble layer in the test? 
`self.harness.model.unit.get_container("magma-orc8r-certifier").add_layer(rest of the stuff)`
At first glance it was quite confusing to see `pebble_ready` being called twice with the `WaitingStatus` being set in the middle.

ghislainbourgeois: Thanks! At first, I tried to look for a method directly in harness to do exactly that and could not find it. I agree that it will be a lot more clear.

ghislainbourgeois: Done, thanks again!

Gmerold (approved):

PR #147, opened 2022-12-27
Fix: updating root certs in the relation data

Gmerold:
# Description

Fixes [TELCO-358](https://warthogs.atlassian.net/browse/TELCO-358)
Updates root certs (`rootCA.pem` and `controller.crt`) in the relation data after they are recreated.

Additionally:
Implements passing `nms-magmalte` k8s service details through the relation data bag
Removed testing.SIMULATE_CAN_CONNECT from unit tests

# Checklist:

- [x] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that validate the behaviour of the software
- [x] I validated that new and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

[TELCO-358]: https://warthogs.atlassian.net/browse/TELCO-358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #156, opened 2023-01-23
Update nginx config file on config changed

ghislainbourgeois:
# Description

Previously, the nginx configuration file was only generated `on_install`. This changes recreates the file `on_config_changed` before restarting the service.

## Checklist

- [x] I have added tests that validate the behaviour of the software.
- [x] I validated that new and existing unit tests pass locally with my changes.
- [x] Any dependent changes have been merged and published in downstream modules.

### DIFF
--- a/orc8r-nginx-operator/src/charm.py
+++ b/orc8r-nginx-operator/src/charm.py
@@ -82,7 +82,7 @@ class MagmaOrc8rNginxCharm(CharmBase):
         )
 
         self.framework.observe(self.on.install, self._on_install)
-        self.framework.observe(self.on.config_changed, self._configure_magma_orc8r_nginx)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(
             self.on.magma_orc8r_nginx_pebble_ready, self._configure_magma_orc8r_nginx
         )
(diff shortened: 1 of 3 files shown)


orc8r-nginx-operator/src/charm.py
```diff
@@ -130,6 +130,21 @@ def _on_install(self, event: InstallEvent) -> None:
         self._generate_nginx_config()
         self._create_additional_orc8r_nginx_services()
 
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
```
Gmerold, line 132: Do you think it would make sense to also check whether the workload container can be connected to or not?

Gmerold (approved):

orc8r-nginx-operator/src/charm.py
```diff
@@ -130,6 +130,21 @@ def _on_install(self, event: InstallEvent) -> None:
         self._generate_nginx_config()
```
gruyaume: Do we need the `_on_install` event handler at all in this case? It all could be handled on config changed.

Gmerold: In [this PR](https://github.com/canonical/charmed-magma-orchestrator/pull/147) we'll install `procps` in `_on_install`.

gruyaume: Ahhh ok but in all cases  we don't need to run `_generate_nginx_config` both on install and on config changed. This means that it'd be run twice when we deploy a given charm.

orc8r-nginx-operator/src/charm.py
ghislainbourgeois: Yes, good catch! Fixed

orc8r-nginx-operator/src/charm.py
ghislainbourgeois: I removed `_generate_nginx_config` from `_on_install`.

gruyaume (approved):

PR #163, opened 2023-01-31
Removing FeG components prior to releasing charmed magma 1.6

gruyaume:
# Description

Charmed Magma Release 1.6 will not support federation gateway, hence removing traces of it for maintenance/supportability reasons

## Checklist


saltiyazan:
Are we sure we want to merge this PR into main, or should it be merged into 1.6?

gruyaume:
> Are we sure we want to merge this PR into main, or should it be merged into 1.6?

Humm well the 1.6 branch will be based on master once we agree it's ready right? Also FeG is not used in main. I think we should add the FeG components if/when we end up supporting FeG.

Edit. Ok I'll keep it there following the mid cycle discussion.

saltiyazan:
> > Are we sure we want to merge this PR into main, or should it be merged into 1.6?
> 
> Humm well the 1.6 branch will be based on master once we agree it's ready right? Also FeG is not used in main. I think we should add the FeG components if/when we end up supporting FeG.

Yes 1.6 will be based on master, but it's not clear to me why we should lose the code in master too. (I'm okay if that is intended :) )

PR #164, opened 2023-01-31
Moves orc8r-libs to their own project

gruyaume:
# Description

Removes libs from the charmed-magma-orchestrator repo and moves them to their own repo (https://github.com/canonical/magma-orc8r-libs)

## Checklist


saltiyazan (approved):

saltiyazan:
Approved, but why is it needed? Won't we have release branches for the libs?

gruyaume:
> Approved, but why is it needed? Won't we have release branches for the libs?

Those libs don't need to follow the same release cycle as the charms that use them. For example, it's possible that the libs are the same for 1.6.

Gmerold (approved):

saltiyazan:
> > Approved, but why is it needed? Won't we have release branches for the libs?
> 
> Those libs don't need to follow the same release cycle as the charms that use them. For example, it's possible that the libs are the same for 1.6.

Alright :)

PR #177, opened 2023-02-08
[Bug Fix] Orc8r-nginx crashes when certifier certificates are ready before nginx.conf file is created

gruyaume:
# Description

Fixes bug where if certifier certificate was ready before the config file was created, the charm would crash

## Checklist

### DIFF
--- a/orc8r-nginx-operator/src/charm.py
+++ b/orc8r-nginx-operator/src/charm.py
@@ -32,7 +32,6 @@ from lightkube.resources.core_v1 import Service
 from ops.charm import (
     CharmBase,
     ConfigChangedEvent,
-    InstallEvent,
     PebbleReadyEvent,
     RelationBrokenEvent,
     RelationJoinedEvent,
@@ -55,6 +54,7 @@ class MagmaOrc8rNginxCharm(CharmBase):
     """An instance of this object everytime an event occurs."""
 
     BASE_CERTS_PATH = "/var/opt/magma/certs"
+    CONFIG_PATH = "/etc/nginx"
     REQUIRED_RELATIONS = ["cert-certifier", "cert-controller"]
     REQUIRED_MAGMA_SERVICES_RELATIONS = ["magma-orc8r-bootstrapper", "magma-orc8r-obsidian"]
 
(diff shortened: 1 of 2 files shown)


ghislainbourgeois (approved):

saltiyazan (approved):

orc8r-nginx-operator/src/charm.py
```diff
@@ -174,6 +174,10 @@ def _configure_magma_orc8r_nginx(
             self.unit.status = WaitingStatus("Waiting for certificates to be available.")
             event.defer()
             return
+        if not self._nginx_config_is_generated:
+            self.unit.status = WaitingStatus("Waiting for nginx config to be generated.")
+            event.defer()
+            return
```
Gmerold, line 171: How about instead of adding this check, we move the content of `_on_install` here?
`_generate_nginx_config` needs to be called on a config changed anyway and `_create_additional_orc8r_nginx_services` is already smart enough to only create services when needed.
Are there any advantages of starting the config `on.install`?

gruyaume: The reason why it's only on install is that this action needs only to be done once. But ultimately yes we can move it here if we think it's better.

Gmerold: I mean I understand why we use `on.install` ;) It's just that `_generate_nginx_config` already is called more than once anyway and the `_create_additional_orc8r_nginx_services` has its own safety mechanism preventing it from creating services over and over again.
I just thought removing `_on_install` could simplify the charm.

gruyaume: I just realised that the nginx config file was created on `config_changed` so indeed I removed it from on install. That being said, I kept the `create_additional_orc8r_nginx_services` in the `on_install` because I think it's too much to make API calls at each Juju event to check whether those services were created when we know that they'll only be created once, whatever happens with the charm.

Let me know what you think.

Gmerold: Works for me.

Gmerold (approved):

PR #228, opened 2023-03-13
Telco 309 upgrade integration tests proposal

Gmerold:
# Description

Drafts itests for the upgrade procedure.

Context:
The approach used in this draft leverages the fact, that the charm installed directly from charmhub uses a container image uploaded to charmhub's container registry, whereas charm installed from a local file uses a container image specified in the `metadata.yaml`. This fact can be used to check whether the charm under test was actually upgraded.
The test scenario deploys the charm from charmhub and validates that the workload container's image actually comes from charmhub. Next, charm under test is built from sources and upgrade/refresh is triggered. Once operation is completed (Juju application is in active/idle again), test validates that the workload container's image has changed to the one specified in charm's `metadata.yaml`.
The advantage of this approach is that it'll work across all the future releases. 
The disadvantage is that it's just a super basic test, which definitely isn't enough to ensure that the upgrade works as expected.

Alternatively we could be testing specific changes between the releases (like configs, startup commands, ports, etc.), but that approach would require maintenance between the releases.

## Checklist

### DIFF
--- a/orc8r-tenants-operator/tests/integration/test_upgrade.py
+++ b/orc8r-tenants-operator/tests/integration/test_upgrade.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+from pathlib import Path
+
+import pytest
+import yaml
+
+logger = logging.getLogger(__name__)
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+
+APPLICATION_NAME = "orc8r-tenants"
+CHARM_NAME = "magma-orc8r-tenants"
+CHARM_RESOURCES = {
+    f"{CHARM_NAME}-image": METADATA["resources"][f"{CHARM_NAME}-image"]["upstream-source"],
+}
+
+
+class TestOrc8rTenants:
+    @pytest.fixture(scope="module")
+    @pytest.mark.abort_on_fail
+    async def setup(self, ops_test):
+        await self._deploy_postgresql(ops_test)
+        await self._deploy_orc8r_tenants_from_latest_stable_channel(ops_test)
+        await self._relate_orc8r_tenants_and_postgresql(ops_test)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=600)
+
+    @pytest.mark.abort_on_fail
+    async def test_workload_container_image_comes_from_charmhub(self, ops_test, setup):
+        image = await self._get_container_image(ops_test, APPLICATION_NAME)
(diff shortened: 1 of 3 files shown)


gruyaume:
Do you mind adding a bit more information and context to the description to make the review easier.

orc8r-tenants-operator/tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
```
gruyaume, line 2: 2021 !!

orc8r-tenants-operator/tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,95 @@
...
+    @pytest.fixture(scope="module")
+    @pytest.mark.abort_on_fail
+    async def setup(self, ops_test):
+        await self._deploy_postgresql(ops_test)
+        await self._deploy_orc8r_tenants_from_latest_stable_channel(ops_test)
+        await self._relate_orc8r_tenants_and_postgresql(ops_test)
+        await ops_test.model.wait_for_idle(apps=[APPLICATION_NAME], status="active", timeout=600)
+
+    @pytest.mark.abort_on_fail
+    async def test_workload_container_image_comes_from_charmhub(self, ops_test, setup):
```
gruyaume, line 31: Why would we want to validate this?

orc8r-tenants-operator/tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+    @staticmethod
+    async def _relate_orc8r_tenants_and_postgresql(ops_test):
+        await ops_test.model.add_relation(
+            relation1=APPLICATION_NAME, relation2="postgresql-k8s:db"
+        )
+
+    @staticmethod
+    async def _get_container_image(ops_test, app_name):
```
gruyaume, line 75: Add typing/return-typing

gruyaume (commented):
Minor comments in the context of trying to understand the thing

orc8r-tenants-operator/tests/integration/test_upgrade.py
Gmerold: I just added it to ensure that the first deployment was actually done from charmhub. That's the assumption for this whole test suite basically. 
It's not necessary, but I thought that if, at some point, the behavior of the deployment from charmhub changes (i.e. it won't be using built-in container registry), we'd like to know about it. This test will start failing if anything like that happens.
Alternatively, we could check that the container image's URL is different than the one specified in the `metadata.yaml`. That would secure the test from failing if charmhub's registry URL changes.

PR #229, opened 2023-03-13
Use nginx rock for orc8r-nginx service

ghislainbourgeois:
# Description

Use nginx rock for orc8r-nginx service. This also change the generation of the configuration file, as upstream Magma container has a script to generate it from a template already inside the container image.

## Checklist

- [x] I have made corresponding changes to the documentation.
- [x] I have added tests that validate the behaviour of the software.
- [x] I validated that new and existing unit tests pass locally with my changes.

### DIFF
--- a/orc8r-nginx-operator/src/nginx.conf.j2
+++ b/orc8r-nginx-operator/src/nginx.conf.j2
@@ -1,0 +1,141 @@
+user root;
+worker_processes auto;
+pid /run/nginx.pid;
+
+events {
+  worker_connections 1024;
+}
+
+http {
+  # Custom JSON-formatted log
+  log_format json_custom escape=json
+    '{'
+      '"nginx.time_local": "$time_local",'
+      '"nginx.remote_addr": "$remote_addr",'
+      '"nginx.request": "$request",'
+      '"nginx.request_method": "$request_method",'
+      '"nginx.request_uri": "$request_uri",'
+      '"nginx.status": $status,'
+      '"nginx.body_bytes_sent": $body_bytes_sent,'
+      '"nginx.request_length": $request_length,'
+      '"nginx.request_time": $request_time,'
+      '"nginx.server_name": "$server_name",'
+      '"nginx.client_serial": "$ssl_client_serial",'
+      '"nginx.client_cn": "$ssl_client_s_dn_cn"'
+    '}';
+
+  ## Fix - [emerg]: could not build the map_hash, you should increase
+  map_hash_bucket_size 64;
+
+  # See https://kubernetes.github.io/ingress-nginx/examples/grpc/#notes-on-using-responserequest-streams
+  grpc_send_timeout 1200s;
+  grpc_read_timeout 1200s;
+  client_body_timeout 1200s;
+
+  # Use a regex to pull the client cert common name out of the DN
+  # The DN will look something like "CN=foobar,OU=,O=,C=US"
+  map $ssl_client_s_dn $ssl_client_s_dn_cn {
+    default "";
+    ~CN=(?<CN>[^/,]+) $CN;
+  }
+
+  # Server block for controller
+  server {
+    listen              8443 ssl http2;
+    server_name         ~^(?<srv>.+)-{{ controller_hostname }}$;
+
+    error_log  /var/log/nginx/error.log info;
+    access_log /var/log/nginx/access.log json_custom;
+
+    ssl_certificate     {{ base_certs_path }}/controller.crt;
+    ssl_certificate_key {{ base_certs_path }}/controller.key;
+    ssl_verify_client on;
+    ssl_client_certificate {{ base_certs_path }}/certifier.pem;
+
+    # Max allowed size for client requests body
+    client_max_body_size 50M;
+
+    location / {
+      resolver {{ resolver }};
+
+      # certifier is internal only service
+      if ($srv = "certifier") {
+        return 403;
+      }
+      # Helm services don't allow for underscores. Magma convention is
+      # to use underscores in service names, so convert any hyphens in the
+      # k8s service name.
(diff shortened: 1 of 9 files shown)


orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -466,25 +460,18 @@ def test_given_container_can_be_connected_when_rootca_certificate_available_then
             path="/var/opt/magma/certs/rootCA.pem", source=test_rootca_cert
         )
 
-    @patch("ops.model.Container.exec")
+    @patch("ops.model.Container.push")
     def test_given_valid_domain_config_set_when_config_changed_then_nginx_config_file_is_created(
```
saltiyazan:
```suggestion
    def test_given_valid_domain_config_set_when_config_changed_then_nginx_config_file_is_pushed(
```
netpicking, but maybe it's better to justify the usage of `patched_push` here :)

ghislainbourgeois: Done, thanks!

orc8r-nginx-operator/tests/unit/test_charm.py
```diff
@@ -138,7 +140,7 @@ def test_given_no_relations_created_when_pebble_ready_event_emitted_then_status_
         )
 
     @patch("ops.model.Container.exists")
-    @patch("ops.model.Container.exec", Mock())
+    @patch("ops.model.Container.exec", new=Mock)
```
gruyaume, line 143: I am still confused about this, I know `new_callable` expects a callable but why would `new` require one as well?

orc8r-nginx-operator/src/nginx.conf.j2
```diff
@@ -0,0 +1,141 @@
...
+  # Use a regex to pull the client cert common name out of the DN
+  # The DN will look something like "CN=foobar,OU=,O=,C=US"
+  map $ssl_client_s_dn $ssl_client_s_dn_cn {
+    default "";
+    ~CN=(?<CN>[^/,]+) $CN;
+  }
+
+  # Server block for controller
+  server {
+    listen              8443 ssl http2;
```
gruyaume: We may want to have a variable for this port since it has to match with the one in the k8s service patch

orc8r-nginx-operator/src/nginx.conf.j2
```diff
@@ -0,0 +1,141 @@
...
+      grpc_set_header Host $srv-orc8r-$k8s_svc.{{ backend }}:9180;
+
+      grpc_set_header x-magma-client-cert-cn $ssl_client_s_dn_cn;
+      grpc_set_header x-magma-client-cert-serial $ssl_client_serial;
+    }
+  }
+
+  # Server block for bootstrapper and any other non-clientcert services
+  server {
+    listen 8444 ssl http2;
```
gruyaume: Same comment here

orc8r-nginx-operator/src/nginx.conf.j2
```diff
@@ -0,0 +1,141 @@
...
+      if ($srv ~* "(\w+)[_](\w+)") {
+        set $srv "$1-$2";
+      }
+      grpc_pass grpc://orc8r-bootstrapper.{{ backend }}:9180;
+    }
+  }
+
+  # Catch-all server block for REST HTTP/1.1 requests from browsers
+  server {
+    listen 9443 ssl default_server;
```
gruyaume: Same comment here

orc8r-nginx-operator/tests/unit/test_charm.py
ghislainbourgeois: The documentation is really confusing for this, it goes on and on about what happens if you omit `new`, but does not really say what it will do if you specify it.

I just wrote a test script to compare what happens in different cases, and I think I understand better what they do.

When using `new`, the target will be replaced by what is provided to `new`. When the target is a class, we should pass a class. When the target is a function, we need to pass a callable (simplified: a function, or an object with a `__call__` method).

When using `new_callable`, the target will be replaced by the result of calling what is provided to `new`.

So, with `new`, if we pass a class, the target will be replaced by the class, and the code could then instantiate it. If we pass an instance of a class, the target is replaced by that instance. If the code expects a class to instantiate, that would break unless it is a callable.

With `new_callable`, if we pass a class, the target will be replaced by an instance of that class. If we pass a function that returns "pizza", the target will be replaced by the string "pizza".

Now, we get into `Mock`. `Mock` is a class, that when instantiated returns an object that:

1. Is a callable. When called, it will return a new instance of `Mock`. ie: `Mock()() -> Mock`
2. Will return a named `Mock` for any attributes accessed. ie: `Mock().a -> Mock(name="mock.a")`

Conclusion, in the case of `Mock`, it does not change much, because of all the magic it does. If you hand-code mocks however, the distinction becomes important.

orc8r-nginx-operator/src/nginx.conf.j2
ghislainbourgeois: Good point, fixed

orc8r-nginx-operator/src/nginx.conf.j2
ghislainbourgeois: Good point, fixed

orc8r-nginx-operator/src/nginx.conf.j2
ghislainbourgeois: Good point, fixed

orc8r-nginx-operator/tests/unit/test_charm.py
gruyaume: Thanks for the answer, it's appreciated, I'll probably come back to it.

gruyaume (approved):

PR #240, opened 2023-04-11
Removes feg-related charms

sanchezfdezjavier:
# Description

Removes veg-related charms.

## Checklist


saltiyazan:
Do we want to remove these charms from edge too?

sanchezfdezjavier:
> Do we want to remove these charms from edge too?

I think we do. There's no plan to tackle feg any time soon, and for doing changes in which you have to grep globally across the repo they are a bit of a pain. Besides, we can always bring those back if something happens and we want to go ahead with feg. However, I'd be surprised if that happens.

gruyaume (approved):

